PR #1, opened 2021-08-24
feat: NFT support for Pool contract

manuelpolzhofer:
Here is my PR to support NFT streaming in the funding pool.

**Pool Contract**
- all internal methods have a underscore
- introduction of `address id` replace all internal usages of msg.sender with `id`
- kept all public view method interfaces which use msg.sender
   - we can break the interface and require `address id` param or a separate method
- no proxy support for NFT because we want to remove it

**NFT Pool Contract**
- unique id based on the `keccak256(abi.encodePacked(nftRegistry, tokenId)` 
- collect and updateSender requires the NFT ownership
   - we might should consider renaming the function to `updateNFTSender`
   
**Tests**
- added first basic tests for NFT streaming
   - NFT to address streaming
   - NFT to NFT streaming
   - address to NFT streaming
- the old javascript tests are still required to be migrated

### DIFF
--- a/src/NFTPool.sol
+++ b/src/NFTPool.sol
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.6;
+pragma experimental ABIEncoderV2;
+
+import {DaiPool, Erc20Pool, ReceiverWeight, Dai} from "./Pool.sol";
+import {IERC721} from "openzeppelin-contracts/token/ERC721/IERC721.sol";
+
+/// @notice NFT pool contract to support streaming based on NFT ownership
+/// A NFT can be a sender or a receiver, a unique id is generated based on
+/// NFT registry address and the tokenId
+contract NFTPool is Erc20Pool {
+    modifier nftOwner (address nftRegistry, uint tokenId) {
+        require(IERC721(nftRegistry).ownerOf(tokenId) == msg.sender, "not-NFT-owner"); _;
+    }
+    constructor(uint64 cycleSecs, Dai dai) Erc20Pool(cycleSecs, dai) {}
+
+    /// @notice generates a unique 20 bytes by hashing the nft registry  and tokenId
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function nftID(address nftRegistry, uint128 tokenId) public pure returns (address id) {
+        // gas optimized without local variables
+        return address(uint160(uint256(
+                keccak256(abi.encodePacked(nftRegistry, tokenId)
+                ))));
+    }
+
+    /// @notice collect the funds of an NFT. Requires the msg.sender to own the NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function collect(address nftRegistry, uint128 tokenId) public nftOwner(nftRegistry, tokenId) {
+        uint128 collected = _collectInternal(nftID(nftRegistry, tokenId));
+        if (collected > 0) {
+            // msg.sender === nft owner
+            _transferToSender(msg.sender, collected);
+        }
+        emit Collected(msg.sender, collected);
+    }
+
+    /// @notice updateSender based on the ownership of an NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function updateSender(
+        address nftRegistry,
+        uint128 tokenId,
+        uint128 topUpAmt,
+        uint128 withdraw,
+        uint128 amtPerSec,
+        ReceiverWeight[] calldata updatedReceivers,
+        ReceiverWeight[] calldata updatedProxies
+    ) public nftOwner(nftRegistry, tokenId) {
+        // msg.sender === nft owner
+        _transferToContract(msg.sender, topUpAmt);
+        uint128 withdrawn =
+        _updateSenderInternal(nftID(nftRegistry, tokenId), topUpAmt, withdraw, amtPerSec,
+            updatedReceivers, updatedProxies);
+        _transferToSender(msg.sender, withdrawn);
+    }
+}
--- a/src/NFTPool.sol
+++ b/src/NFTPool.sol
@@ -3,17 +3,18 @@
 pragma solidity ^0.8.6;
 pragma experimental ABIEncoderV2;
 
-import {DaiPool, Erc20Pool, ReceiverWeight, Dai} from "./Pool.sol";
+import {DaiPool, ReceiverWeight, Dai} from "./Pool.sol";
 import {IERC721} from "openzeppelin-contracts/token/ERC721/IERC721.sol";
 
 /// @notice NFT pool contract to support streaming based on NFT ownership
 /// A NFT can be a sender or a receiver, a unique id is generated based on
 /// NFT registry address and the tokenId
-contract NFTPool is Erc20Pool {
+contract NFTPool is DaiPool {
     modifier nftOwner (address nftRegistry, uint tokenId) {
-        require(IERC721(nftRegistry).ownerOf(tokenId) == msg.sender, "not-NFT-owner"); _;
+        require(IERC721(nftRegistry).ownerOf(tokenId) == msg.sender, "not-NFT-owner");
+        _;
     }
-    constructor(uint64 cycleSecs, Dai dai) Erc20Pool(cycleSecs, dai) {}
+    constructor(uint64 cycleSecs, Dai dai) DaiPool(cycleSecs, dai) {}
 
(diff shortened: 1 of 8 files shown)


contracts/NFTPool.sol
```diff
@@ -0,0 +1,59 @@
...
+    /// @notice collect the funds of an NFT. Requires the msg.sender to own the NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function collect(address nftRegistry, uint128 tokenId) public nftOwner(nftRegistry, tokenId) {
+        uint128 collected = _collectInternal(nftID(nftRegistry, tokenId));
+        if (collected > 0) {
+            // msg.sender === nft owner
+            _transferToSender(msg.sender, collected);
+        }
+        emit Collected(msg.sender, collected);
```
manuelpolzhofer, line 37: Ah, I think here the event should be the unique `nftID` and not the `msg.sender` who collected for that NFT

contracts/NFTPool.sol
```diff
@@ -0,0 +1,59 @@
...
+        // gas optimized without local variables
+        return address(uint160(uint256(
+                keccak256(abi.encodePacked(nftRegistry, tokenId)
+                ))));
+    }
+
+    /// @notice collect the funds of an NFT. Requires the msg.sender to own the NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function collect(address nftRegistry, uint128 tokenId) public nftOwner(nftRegistry, tokenId) {
```
CodeSandwich, line 31: What's the use case for streaming funds to an NFT?

contracts/NFTPool.sol
```diff
@@ -0,0 +1,59 @@
...
+    modifier nftOwner (address nftRegistry, uint tokenId) {
+        require(IERC721(nftRegistry).ownerOf(tokenId) == msg.sender, "not-NFT-owner"); _;
+    }
+    constructor(uint64 cycleSecs, Dai dai) Erc20Pool(cycleSecs, dai) {}
+
+    /// @notice generates a unique 20 bytes by hashing the nft registry  and tokenId
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function nftID(address nftRegistry, uint128 tokenId) public pure returns (address id) {
+        // gas optimized without local variables
```
CodeSandwich, line 22: I didn't know that trick! What does it do? Does it skip memory allocation when passing data from `encodePacked` to `keccak256`?

BTW integer casting of local variables in Solidity is a noop in EVM, they are always uint256, it only changes how they're handled by Solidity :)

contracts/NFTPool.sol
```diff
@@ -0,0 +1,59 @@
...
+pragma solidity ^0.7.5;
+pragma experimental ABIEncoderV2;
+
+import {DaiPool, Erc20Pool, ReceiverWeight, Dai} from "./Pool.sol";
+import {IERC721} from "openzeppelin-contracts/token/ERC721/IERC721.sol";
+
+/// @notice NFT pool contract to support streaming based on NFT ownership
+/// A NFT can be a sender or a receiver, a unique id is generated based on
+/// NFT registry address and the tokenId
+contract NFTPool is Erc20Pool {
```
CodeSandwich, line 12: If it works on Dai, shouldn't it inherit `DaiPool`? We'd get support for quasi-ERC-712 signatures for free.

contracts/NFTPool.sol
```diff
@@ -0,0 +1,59 @@
...
+            _transferToSender(msg.sender, collected);
+        }
+        emit Collected(msg.sender, collected);
+    }
+
+    /// @notice updateSender based on the ownership of an NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function updateSender(
+        address nftRegistry,
```
CodeSandwich, line 44: What's the use case for supporting multiple NFT registries? If we're going to mint and burn NFTs based on support, it's probably going to be a single registry managed by the pool (or even the pool itself).

CodeSandwich (commented):
That's a lot of good work done here :+1: 

In the future would it be a problem for you to work on feature branches in the original `radicle-dev` repo? It's a bit easier to cooperate when everything is in a single place.

contracts/NFTPool.sol
manuelpolzhofer: The current idea is that every Radicle Funding project would have their own NFT registry but we will only have one Pool contract deployed. 

One open questions is still if we want to whitelist the NFT registries which are allowed to use the contract.

Otherwise, potential malicious NFT registries could try to exploit the system

contracts/NFTPool.sol
manuelpolzhofer: Yes, I will add support for that as well together with the DAI `permit`. I already tried but sadly it resulted in too many local variables

contracts/NFTPool.sol
manuelpolzhofer: Currently, not a use-case for Radicle Funding it could be useful as an additional abstraction layer to easily transfer your streams

contracts/NFTPool.sol
manuelpolzhofer: Ah no, I just added to comment because in a normal programming language you might would make some local variables to increase readability. 

```
bytes32 hash = ...;
address id = ...;
return id;
```

Yeah, the casting looks a bit ugly but you can't cast bytes32 to uint160 directly. We might can do it in  `assembly`

contracts/NFTPool.sol
CodeSandwich: > every Radicle Funding project would have their own NFT registry

That means that as a sender I'd have as many "identities" and sender accounts to top up, as many projects I'd want to support :thinking:

contracts/NFTPool.sol
CodeSandwich: Casting is fine, I just thought that the optimizer is smart enough to not create the unneeded local variables. That would leave us more freedom with formatting, as you've shown.

contracts/NFTPool.sol
CodeSandwich: What do you mean? Have you hit a limitation of the arguments? Don't worry about gas price of making a call with such a long list of arguments, shaving off a single separate call to `approve` will immediately make it up for the user.

src/NFTPool.sol
```diff
@@ -0,0 +1,74 @@
...
+    /// @notice updateSender based on the ownership of an NFT
+    /// @param nftRegistry address of the NFT specific registry
+    /// @param tokenId the unique token id for the NFT registry
+    function updateSender(
+        address nftRegistry,
+        uint128 tokenId,
+        uint128 topUpAmt,
+        uint128 withdraw,
+        uint128 amtPerSec,
+        ReceiverWeight[] calldata updatedReceivers,
```
CodeSandwich, line 66: How to send funds to an NFT receiver? We probably should ban sending to a raw address, because it will make rotation of keys and transfer of an NFT to the new one useless.

contracts/NFTPool.sol
manuelpolzhofer: For me the NFT is not an identity more the support itself.  

We want a single NFT for each project you support because of the influence, access and badge aspect. 

We were discussion it in the past and current the idea would be that you have to topUp multiple NFTs technically. From the UX perspective we could bundle multiple topUps in on transaction, etc. We are aware that this is currently one downside of the NFT approach.

I know that is a bit against the current benefit of the funding pool where i only need to topUp once for multiple receivers.

Theoretically we could use the users address to stream towards your NFTs and they are streaming to towards the projects. (but then we need to ensure that `collect` is called and the user needs to be aware of the implications if he would sell the NFT).


From the current requirement perspective: we have the following connection: a NFT should be inactive for influence and access token if the streaming stops.

We have two options for alternative designs:
- we could say we always check the owner of the NFT if the owner (address) is still streaming to the project). 
   - if not, access and influence is inactive
- Here we have the following problems, what should happen if the NFT get transfered? We could overwrite the transferFrom method and adjust the deltas of the receiver and previous owner. (however we would have to prevent airdrops and the receiver needs to accept the NFT first with some kind of flag)

The other option would be each NFT has a supporter address which is independent of the NFT owner. However, here we would also have to discuss the behaviour in case of an transfer.

src/NFTPool.sol
manuelpolzhofer: you  would send to `address id = hash(registry, address)`

Inside the collect function I am doing the following:

```
// requires the address to own the nft
_collectInternal(nftID(nftRegistry, tokenId));
```
In the receiverWeights array it is basically abstracted. You don't know if the address an NFT id or a normal address

CodeSandwich (approved):
:+1:

PR #2, opened 2021-08-25
Remove pool proxy

CodeSandwich:
This is the first step to implement drips. Also copies docs for how pool works.


src/NFTPool.sol
```diff
@@ -63,11 +61,9 @@ contract NFTPool is DaiPool {
         uint128 topUpAmt,
         uint128 withdraw,
         uint128 amtPerSec,
-        ReceiverWeight[] calldata updatedReceivers,
-        ReceiverWeight[] calldata updatedProxies
+        ReceiverWeight[] calldata updatedReceivers
     ) public nftOwner(nftRegistry, tokenId) {
-        _sendFromNFT(nftID(nftRegistry, tokenId),
-            topUpAmt, withdraw, amtPerSec, updatedReceivers, updatedProxies);
+        _sendFromNFT(nftID(nftRegistry, tokenId), topUpAmt, withdraw, amtPerSec, updatedReceivers);
```
manuelpolzhofer, line 66: Ah, nice already up to date with master

src/Pool.sol
```diff
@@ -688,81 +451,7 @@ abstract contract Pool {
...
-            for (uint32 i = 0; i < receiversCount; i++) {
-                Receiver storage receiver = receivers[receiversList[i].receiver];
-                int128 weight = int128(uint128(receiversList[i].weight));
-                receiver.amtDeltas[currCycle].thisCycle -= totalDelta * weight;
-            }
-        }
-    }
-
-    function currTimestamp() public view returns (uint64) {
+    function _currTimestamp() internal view returns (uint64) {
```
manuelpolzhofer, line 454: Why not public if it is a `view`?

src/libraries/ReceiverWeights.sol
```diff
@@ -79,65 +74,36 @@ library ReceiverWeightsImpl {
...
-        self.attachToList(proxy);
-        previousWeight = self.data[proxy].weightProxy;
-        self.data[proxy].weightProxy = weight;
-    }
-
-    /// @notice Ensures that the weight for a specific receiver is attached to the list
-    /// @param receiver The receiver whose weight should be attached
-    function attachToList(ReceiverWeights storage self, address receiver) internal {
         require(receiver != ADDR_ROOT, "Invalid receiver address");
+        // Ensure that the weight for a specific receiver is attached to the list
```
manuelpolzhofer, line 99: Ah okay you removed the `attachToList` because it is only used once and we can save some gas.

src/libraries/ReceiverWeights.sol
```diff
@@ -79,65 +74,36 @@ library ReceiverWeightsImpl {
...
-    function isZeroed(ReceiverWeights storage self) internal view returns (bool) {
-        return self.data[ADDR_ROOT].next == ADDR_ROOT;
-    }
-
     /// @notice Set weight for a specific receiver
     /// @param receiver The receiver to set weight
     /// @param weight The weight to set
     /// @return previousWeight The previously set weight, may be zero
-    function setReceiverWeight(
+    function setWeight(
```
manuelpolzhofer, line 93:
```suggestion
    function _setWeight(
```

Maybe lets change it for consistency in the library as well.

src/Pool.sol
```diff
@@ -688,81 +451,7 @@ abstract contract Pool {
         amtDeltas[thisCycle].nextCycle += int128(uint128(nextCycleSecs)) * amtPerSecDelta;
     }
 
-    /// @notice Stops proxy of `msg.sender` for the duration of the modified function.
-    /// This removes and then restores any effects of the proxy on all of its receivers' futures.
-    /// It allows the function to safely modify any properties of the proxy
-    /// without having to updating the state of its receivers.
-    modifier suspendProxy {
```
manuelpolzhofer, line 695: Ah nice, that this is removed. In Solidity it is anyway kind of standard to use modifiers only for access-control

src/Pool.sol
```diff
@@ -224,7 +154,7 @@ abstract contract Pool {
         Receiver storage receiver = receivers[msg.sender];
         uint64 collectedCycle = receiver.nextCollectedCycle;
         if (collectedCycle == 0) return 0;
-        uint64 currFinishedCycle = currTimestamp() / cycleSecs;
+        uint64 currFinishedCycle = _currTimestamp() / cycleSecs;
```
manuelpolzhofer, line 157: I think it might be nice to have individual functions for the calculate cycle operations based on timestamp operations. Instead of having one for the current timestamp.

manuelpolzhofer (approved):
LGTM.

Very nice. It simplifies the complexity a lot.

src/Pool.sol
CodeSandwich: Because it's not a useful public API member. Who want to use it? It's also much harder to make any changes in a public API.

src/libraries/ReceiverWeights.sol
CodeSandwich: I doubt that it saves any gas, if the solidity optimizer isn't very different from regular compilers, it probably was inlining anyway :shrug:

src/libraries/ReceiverWeights.sol
CodeSandwich: I'm not sure about that, `ReceiverWeights` is a library and all its functions are `internal`. Should every function in every library be prefixed with an underscore?

src/Pool.sol
CodeSandwich: It's a good idea, but I'd rather port the tests before doing any refactoring, what do you think?

PR #5, opened 2021-09-07
Port tests from radicle-contracts

CodeSandwich:


manuelpolzhofer (approved):
LGTM, very nice.

Some general ideas for improvement. I think every number in the test could be with the post fix `number ether`. For having more realistic numbers in the default test.

I think we can make some nice fuzz test based on that.

PR #9, opened 2021-09-07
chore: moved TestDai.sol into test folder

manuelpolzhofer:
-  moved TestDai.sol into test folder
- contract dependencies passed as addresses


src/Pool.sol
```diff
@@ -574,7 +586,7 @@ contract Erc20Pool is Pool {
 contract DaiPool is Erc20Pool {
     // solhint-disable no-empty-blocks
     /// @notice See `Erc20Pool` constructor documentation for more details.
-    constructor(uint64 cycleSecs, Dai dai) Erc20Pool(cycleSecs, dai) {}
+    constructor(uint64 cycleSecs, address dai) Erc20Pool(cycleSecs, dai) {}
```
manuelpolzhofer, line 589: Ah maybe here I should have kept it to `IDai`.  Because the interface check for the `permit method would happen the first time in the updateSenderAndPermit method.

CodeSandwich: Definitely it should be IDai. Then the constructor for Erc20Pool can be reverted to accepting IERC20s.

src/test/Pool.t.sol
```diff
@@ -20,7 +20,7 @@ contract PoolTest is BaseTest {
         emit log_named_uint("block.timestamp start", block.timestamp);
 
         dai = new Dai();
-        pool = new NFTPool(CYCLE_SECS, dai);
+        pool = new NFTPool(CYCLE_SECS, IDai(address(dai)));
```
CodeSandwich, line 23: In `src/test/TestDai.sol` we should have `contract Dai is IDai {`, it'll be more precise and will prevent need for things like this.

manuelpolzhofer: Done

src/Pool.sol
```diff
@@ -19,7 +19,7 @@ interface IDai is IERC20 {
         uint8 v,
         bytes32 r,
         bytes32 s
-    ) external;
+    ) virtual external;
```
CodeSandwich, line 22: Huh, why? Isn't everything in an interface virtual by definition?

src/test/Pool.t.sol
CodeSandwich: Ok, now we can remove these castings, right?

src/Pool.sol
manuelpolzhofer: Ah, yes. The `override` for the permit method was enough. More checks in higher Solidity versions :)

CodeSandwich (approved):
Good job, it's nice and clean now! :broom:

PR #10, opened 2021-09-07
feat: return withdrawn amount from updateSender methods + cleanup tests

manuelpolzhofer:


CodeSandwich (changes requested):
Nice addition, could be handy! Could you document the returned values? It would be great to update the pool tests to check this value too.

manuelpolzhofer:
Yes, I will add comments. Maybe lets merge #8  first and i will add it afterwards to the pool tests.

CodeSandwich (approved):

PR #13, opened 2021-09-08
feat: pool view methods with address id

manuelpolzhofer:


src/Pool.sol
```diff
@@ -160,9 +160,10 @@ abstract contract Pool {
 
     /// @notice Returns amount of received funds available for collection
     /// by the sender of the message
```
CodeSandwich, line 162: This is not true

CodeSandwich (approved):
Nice!

PR #14, opened 2021-09-08
Add publicly callable topUp and collect

CodeSandwich:


src/Pool.sol
```diff
@@ -178,14 +178,14 @@ abstract contract Pool {
         return uint128(collected);
     }
 
-    /// @notice Collects all received funds available for collection
-    /// by the sender of the message and sends them to that sender
-    function collect() public virtual {
-        uint128 collected = _collectInternal(msg.sender);
+    /// @notice Collects all received funds available for the user and sends them to that user
+    /// @param id The id of the user.
+    function collect(address id) public virtual {
```
manuelpolzhofer, line 183: This method would allow an attack on the current NFT design.
We will change it anyway, so it doesn't matter.

However, together with the current NFT based approach I could just calculate your NFT `address id` and call collect.
Your funds would be sent to the `address id` instead of msg.sender.

src/test/Pool.t.sol
```diff
@@ -118,19 +130,27 @@ contract EthPoolTest is DSTest {
     }
 
     function collect(PoolUser user, uint128 expectedAmt) internal {
-        assertCollectable(user, expectedAmt);
-        uint expectedBalance = user.balance() + expectedAmt;
+        collect(user, user, expectedAmt);
```
manuelpolzhofer, line 133: That's elegant, doesn't require to adjust existing tests

manuelpolzhofer (commented):
Looks good. 

Except the potential attack. If you agree here, we maybe should wait with merging until we adopted the new design.

src/Pool.sol
CodeSandwich: Hmm, yes and no. You're right, if one decides to send funds to an ID calculated for an NFT, then the funds can be locked by a malicious 3rd party. But AFAIK streaming to an NFT is not a supported use case. If one wants to stream to an arbitrary address, we can't guarantee that it'll be collectable by anybody.

manuelpolzhofer: I added this method: https://github.com/radicle-dev/radicle-streaming/blob/master/src/NFTPool.sol#L32

Maybe you can remove it. Then I am fine with merging now.

We anyway will implement the streaming NFT proposal.

CodeSandwich: Removed

src/NFTPool.sol
```diff
@@ -26,18 +26,6 @@ contract NFTPool is DaiPool {
                 ))));
     }
 
-    /// @notice collect the funds of an NFT. Requires the msg.sender to own the NFT
-    /// @param nftRegistry address of the NFT specific registry
-    /// @param tokenId the unique token id for the NFT registry
-    function collect(address nftRegistry, uint128 tokenId) public nftOwner(nftRegistry, tokenId) {
```
manuelpolzhofer, line 32: 👍

manuelpolzhofer (approved):

PR #21, opened 2021-09-23
Deduplicate stream update events 

CodeSandwich:
This is the first step toward adding sub-senders. A good side is that because events emission is postponed, now they can be easily deduplicated. Please review thoroughly, dapp tools have no way to test events.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -131,6 +131,17 @@ abstract contract Pool {
         int128 nextCycle;
     }
 
+    struct StreamUpdates {
+        uint256 length;
+        StreamUpdate[] updates;
+    }
+
+    struct StreamUpdate {
+        address receiver;
+        uint128 amtPerSec;
+        uint64 endTime;
+    }
+
     /// @dev Details about all the senders, the key is the owner's address
     mapping(address => Sender) internal senders;
     /// @dev Details about all the receivers, the key is the owner's address
@@ -226,16 +237,26 @@ abstract contract Pool {
         uint128 amtPerSec,
         ReceiverWeight[] calldata updatedReceivers
     ) internal returns (uint128 withdrawn) {
-        _stopSending(id);
+        uint256 maxUpdates = senders[id].weightCount + updatedReceivers.length;
+        StreamUpdates memory updates = StreamUpdates({
+            length: 0,
+            updates: new StreamUpdate[](maxUpdates)
+        });
+
+        _stopSending(id, updates);
         _topUp(id, topUpAmt);
         withdrawn = _withdraw(id, withdrawAmt);
         _setAmtPerSec(id, amtPerSec);
         for (uint256 i = 0; i < updatedReceivers.length; i++) {
             _setReceiver(id, updatedReceivers[i].receiver, updatedReceivers[i].weight);
         }
-        Sender storage sender = senders[id];
-        emit SenderUpdated(id, sender.startBalance, sender.amtPerSec);
-        _startSending(id);
+        _startSending(id, updates);
+
+        emit SenderUpdated(id, senders[id].startBalance, senders[id].amtPerSec);
+        for (uint256 i = 0; i < updates.length; i++) {
+            StreamUpdate memory update = updates.updates[i];
+            emit SenderToReceiverUpdated(id, update.receiver, update.amtPerSec, update.endTime);
+        }
     }
 
     /// @notice Adds the given amount to the senders balance of the user.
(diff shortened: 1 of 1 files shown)


src/Pool.sol
```diff
@@ -419,14 +445,21 @@ abstract contract Pool {
...
-                emit SenderToReceiverUpdated(id, receiverAddr, amtPerSec, timeEnd);
-            } else {
-                // Sending is stopping
-                emit SenderToReceiverUpdated(id, receiverAddr, 0, _currTimestamp());
+
+            // Find an existing receiver stream log to update
+            uint256 updated = 0;
+            while (updated < updates.length && updates.updates[updated].receiver != receiverAddr) {
+                updated++;
+            }
```
manuelpolzhofer, line 453: I understand the code but I still need to understand why we need to deduplicate the events for subSenders.

I will look at the other PRs to understand it and comeback.

Obviously this while loop is not so nice for the performance.


I think there are some potential optimisations:

```solidity
// lengthBefore is the length before starting to iterate the linked list
 while (updated < lengthBefore && updates.updates[updated].receiver != receiverAddr) {
                updated++;
            }
// New receiver, will be pushed
 if (updated == lengthBefore) {
        updates.length++;
        updated = updates.length
 }
            
            
```
The other one:
```solidity
   // counterHit initialized with zero
   uint256 updated = lastHit;
    while (updated < updates.length && updates.updates[updated].receiver != receiverAddr) {
                updated++;
    }
    if(updated < updates.length) {
        lastHit=updated;
    }
```
I think this should work because the order in the array is very similar to the linked list.

In the startSending case the array will only contain the previous elements and the removed elements. (before iteration of the linked-list)

We don't know the position of the first match from the linked-list in the array but we know the nextMatch needs to come after the first Match (updates.updates[updated].receiver == receiverAddr)).

If we have the first match it means in the startSending case we already passed all newly added elements in the linkedList.

Thats my reasoning here. Maybe I miss something.

src/Pool.sol
```diff
@@ -419,14 +445,21 @@ abstract contract Pool {
...
+            // New receiver, will be pushed
+            if (updated == updates.length) {
+                updates.length++;
             }
+            updates.updates[updated] = StreamUpdate({
+                receiver: receiverAddr,
+                amtPerSec: amtPerSecDelta > 0 ? uint128(amtPerSecDelta) : 0,
+                endTime: amtPerSecDelta > 0 ? timeEnd : _currTimestamp()
+            });
```
manuelpolzhofer, line 462: Just for understanding.

Lets  say I have already a stream starting from  
```
t=0
endTime=5
amtPerSec=1
```

Let say I now increase my topUp (enough for t=10 to end and my amtPerSec at  block.timestamp=3.

Currently in the stopSending we would throw an event with:
SenderToReceiverUpdated  amPerSec=1, t=3 (in stopSending)
SenderToReceiverUpdated amPerSec=2 t= 10 (in startSending)

with the new logic we would overwrite the updates array entry and only throw a 
SenderToReceiverUpdated amPerSec=2 t= 10 event.

Correct?

CodeSandwich: Correct. While the current one is technically correct, it's excessive and causes noise, which needs to be handled when processing events.

src/Pool.sol
CodeSandwich: Great ideas! The first optimization does yield up to a few hundred gas per test case. The second is worsening the results, there's probably too much to manage on the stack and memory access is really cheap. One surprisingly good thing to do in terms of gas is reverting to separate paths for stopping and starting sending, I'm not sure why. Pushed.

src/Pool.sol
manuelpolzhofer: Yeah, we should maybe compare the gas differences if it is worth removing the event noise.

src/Pool.sol
manuelpolzhofer: Interesting.

manuelpolzhofer: > One surprisingly good thing to do in terms of gas is reverting to separate paths for stopping and starting sending, I'm not sure why.

Mhm, I think previously you had two IFs in the shorthand form now it is just one.

manuelpolzhofer: Previously we also had the while loop iteration in the stopSending case.

manuelpolzhofer (approved):
looks good. if we want to deduplicate the streaming events

src/Pool.sol
CodeSandwich: > previously you had two IFs in the shorthand form
I've tried both the two ternary operators and the single `if`, both were worse than this.
> Previously we also had the while loop iteration in the stopSending case.
In `stopSending` case it was making 0 iterations, the first check of the first `while` condition check was immediately skipping the loop. To me it should be just as cheap as having an `if` skipping a part of code.

PR #26, opened 2021-09-29
feat: simplified nextWeightPrunning

manuelpolzhofer:
I think it is not required to pass the `prevReceiverHint` as a parameter and `receiverHint` as a return parameter in the **nextWeightPrunning** function.

- param prevReceiverHint
   - It can be read inside the function `address nextReceiverHint = self.data[prevReceiver].next;`
   - i think it is cleaner if only the prevReceiver needs to be passed
   
- returned receiverHint
  - not required anymore

The tests are passing or did I miss anything?

### DIFF
--- a/src/libraries/ReceiverWeights.sol
+++ b/src/libraries/ReceiverWeights.sol
@@ -28,33 +28,34 @@ library ReceiverWeightsImpl {
     /// between the current and the next item from the list.
     /// Iterating over the whole list prunes all the zeroed items.
     /// @param prevReceiver The previously returned `receiver` or ADDR_ROOT to start iterating
-    /// @param prevReceiverHint The previously returned `receiverHint`
     /// or ADDR_ROOT to start iterating
     /// @return receiver The receiver address, ADDR_ROOT if the end of the list was reached
-    /// @return receiverHint A value passed as `prevReceiverHint` on the next call
     /// @return weight The receiver weight
     function nextWeightPruning(
         ReceiverWeights storage self,
-        address prevReceiver,
-        address prevReceiverHint
+        address prevReceiver
     )
         internal
         returns (
             address receiver,
-            address receiverHint,
             uint32 weight
         )
     {
-        if (prevReceiver == ADDR_ROOT) prevReceiverHint = self.data[ADDR_ROOT].next;
-        receiver = prevReceiverHint;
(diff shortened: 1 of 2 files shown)


manuelpolzhofer:
I think the same could be done for `nextWeight` method as well but I will wait for your feedback

CodeSandwich:
That was the original design, it's a clean and simple approach. Unfortunately it means that each item is read twice, first as `self.data[receiver].next` and then as `self.data[prevReceiver].next`. The hint has been added to remove that and cut gas usage: https://github.com/radicle-dev/radicle-contracts/pull/75

manuelpolzhofer:
I think i fixed the unnecessary  `self.data[receiver].next` read now.

 Can you take a second look?

src/libraries/ReceiverWeights.sol
```diff
@@ -28,33 +28,27 @@ library ReceiverWeightsImpl {
...
-            address receiverHint,
-            uint32 weight
-        )
+        returns (address receiver, uint32 weight)
     {
-        if (prevReceiver == ADDR_ROOT) prevReceiverHint = self.data[ADDR_ROOT].next;
-        receiver = prevReceiverHint;
+        // next receiver hint because if it is an zero weight receiver
+        // it will be removed and not the next one
+        address nextReceiverHint = self.data[prevReceiver].next;
```
CodeSandwich, line 40: There's still this line, which guarantees an additional SLOAD per `nextWeightPruning` call

manuelpolzhofer: Let's assume the case of no zero weights.
Each `next` call would only happen once exactly here.

In case of zero weights we would read the next after the zero in the loop.|

I don't see the additional SLOAD?

CodeSandwich: In the case you've described we're first SLOADing cell `self.data[prevReceiver]` to get its `.next` and then we're SLOADing cell `self.data[receiver]` to get its `.weight`. In the current `master` we'd read `self.data[receiver]` to get both `.next` and `.weight`, which is a single SLOAD.

manuelpolzhofer: Ahh, I see. In that way you only have one SLOAD to get `.weight` and `.next` to save gas but you need to return the `.next` and pass it as parameter. 

Nice optimisation.

PR #31, opened 2021-10-11
Add dripsFraction to sender configuration

CodeSandwich:
The first step toward adding drips

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -62,10 +62,12 @@ abstract contract Pool {
     uint64 internal constant MAX_TIMESTAMP = type(uint64).max - 2;
     /// @notice Maximum sum of all receiver weights of a single sender.
     /// Limits loss of per-second funding accuracy, they are always multiples of weights sum.
-    uint32 public constant SENDER_WEIGHTS_SUM_MAX = 10000;
+    uint32 public constant SENDER_WEIGHTS_SUM_MAX = 10_000;
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant SENDER_WEIGHTS_COUNT_MAX = 100;
+    /// @notice Maximum value of drips fraction
+    uint32 public constant DRIPS_FRACTION_MAX = 1_000_000;
     /// @notice The amount passed as the withdraw amount to withdraw all the funds
     uint128 public constant WITHDRAW_ALL = type(uint128).max;
     /// @notice The amount passed as the amount per second to keep the parameter unchanged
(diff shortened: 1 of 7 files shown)


src/Pool.sol
```diff
@@ -453,6 +475,25 @@ abstract contract Pool {
...
+        require(dripsFraction <= DRIPS_FRACTION_MAX, "Drip fraction too high");
+        sender.dripsFraction = dripsFraction;
+    }
+
+    /// @notice Gets the fraction of received funds to be dripped by the provided user.
+    /// @param userAddr The address of the user
+    /// @return dripsFraction The fraction of received funds to be dripped.
+    /// A value from 0 to `DRIPS_FRACTION_MAX` inclusively,
+    /// where 0 means no dripping and `DRIPS_FRACTION_MAX` dripping everything.
+    function getDripsFraction(address userAddr) public view returns (uint32 dripsFraction) {
```
manuelpolzhofer, line 493:
```suggestion
    function dripsFraction(address userAddr) public view returns (uint32 dripsFraction) {
```
In standard Solidity coding style it would be without a `get`

src/Pool.sol
```diff
@@ -66,6 +66,8 @@ abstract contract Pool {
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant SENDER_WEIGHTS_COUNT_MAX = 100;
+    /// @notice Maximum value of drips fraction
+    uint32 public constant DRIPS_FRACTION_MAX = 10000;
```
manuelpolzhofer, line 70: Maybe we should just have on constant for fraction operations?
Currently `SENDER_WEIGHTS_SUM_MAX` has the same value. 

I am also wondering if we want a higher precision in general here. 

Consider the following case: Let say a project receives more the an 1m DAI.
Let say the want to give aways everything above the 1m in drips.

```
Project receives: 1.234.567 DAI
Wants to keep 1m
1-(1000000/1234567)=0.1899994087 (drip factor)
```

I would like to use maybe 10^27 here.

src/Pool.sol
```diff
@@ -453,6 +475,25 @@ abstract contract Pool {
         return subSenders[senderAddr][subSenderId].amtPerSec;
     }
 
+    /// @notice Sets the fraction of received funds to be dripped by the sender.
+    /// @param sender The updated sender
+    /// @param dripsFraction The fraction of received funds to be dripped.
+    /// Must be a value from 0 to `DRIPS_FRACTION_MAX` inclusively,
+    /// where 0 means no dripping and `DRIPS_FRACTION_MAX` dripping everything.
+    function _setDripsFraction(Sender storage sender, uint32 dripsFraction) internal {
```
manuelpolzhofer, line 483: I think if we only use a method once we can also remove it.

src/Pool.sol
CodeSandwich: > Maybe we should just have on constant for fraction operations?

I wouldn't mix them together, they have different consequences and `SENDER_WEIGHTS_SUM_MAX` is low for a reason. The fact that they're equal is a confusing coincidence.

> Let say the want to give aways everything above the 1m in drips.

That's not a use case for drips, you need to declare dripped fraction BEFORE funds are streamed to you, otherwise you could cheat your supporters on drips.

> I would like to use maybe 10^27 here.

I've been thinking about what scale should drips fraction use. The first idea was to just slap `type(uint128).max` and call it a day. But the dripped fraction is declared before you receive anything, so laser precision doesn't seem that useful. The other aspect is that numbers capped at 10^4 are very easy to work with for humans on low level, e.g. in tests or in etherscan. On high level the UI will probably ask the users for percentage on scale 0-100 anyway and multiply it to reach the desired scale. I'm OK with expanding the drips fraction scale, but I don't see the benefits of pushing it very far.

src/Pool.sol
CodeSandwich: Ok, should we keep for now `getDripsFraction` for consistency with other methods? IMO with addition of drips and possible extraction of sender config to calldata we'll need to rename or remove a few methods soon anyway.

manuelpolzhofer: Okay 👍

PR #32, opened 2021-10-11
Drip when collecting

CodeSandwich:
The second step toward adding drips. Depends on https://github.com/radicle-dev/radicle-streaming/pull/31

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -174,11 +174,13 @@ abstract contract Pool {
     }
 
     struct Receiver {
+        // The amount collectable independently from cycles
+        uint128 collectable;
         // The next cycle to be collected
         uint64 nextCollectedCycle;
+        // --- SLOT BOUNDARY
         // The amount of funds received for the last collected cycle.
-        // It never is negative, it's a signed integer only for convenience of casting.
-        int128 lastFundsPerCycle;
+        uint128 lastFundsPerCycle;
         // --- SLOT BOUNDARY
         // The changes of collected amounts on specific cycle.
         // The keys are cycles, each cycle `C` becomes collectable on timestamp `C * cycleSecs`.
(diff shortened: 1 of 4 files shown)


src/Pool.sol
```diff
@@ -213,51 +216,93 @@ abstract contract Pool {
...
+            collected -= dripped;
+            uint128 drippedPerWeight = dripped / sender.weightSum;
+            // Iterating over receivers, see `ReceiverWeights` for details
+            address dripsAddr = ReceiverWeightsImpl.ADDR_ROOT;
+            address hint = ReceiverWeightsImpl.ADDR_ROOT;
+            while (true) {
+                uint32 weight;
+                (dripsAddr, hint, weight) = sender.receiverWeights.nextWeight(dripsAddr, hint);
+                if (dripsAddr == ReceiverWeightsImpl.ADDR_ROOT) break;
+                receivers[dripsAddr].collectable += drippedPerWeight * weight;
```
manuelpolzhofer, line 303: Should  we add some sanity checks? I already checked `weightSum` seems correct, etc.

We could track the `totalDripped` here in the loop and afterwards 
```
require(dripped == totalDdripped)
```

At least for stress testing it, would be maybe nice to have it.

src/Pool.sol
```diff
@@ -213,51 +216,93 @@ abstract contract Pool {
     /// @notice Returns amount of received funds available for collection
     /// @param receiverAddr The address of the receiver
     /// @return collected The available amount
-    function collectable(address receiverAddr) public view returns (uint128) {
+    function collectable(address receiverAddr)
+        public
+        view
+        returns (uint128 collected, uint128 dripped)
```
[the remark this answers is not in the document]
CodeSandwich: Isn't `totalCollected == kept + dripped`? It's not very useful and it adds even more elements to the API with additional rules.

src/Pool.sol
CodeSandwich: I'm a bit skeptical about usefulness of adding `require(dripped == totalDripped)` to production code for sake of stress testing. We must test if everybody actually gets what they're supposed to get and if `dripped != totalDripped`, these checks will catch that and many more issues.

CodeSandwich: Added the sanity check

manuelpolzhofer: Okay nice. However, if we run into a problem and it reverts the fund would be locked.

src/Pool.sol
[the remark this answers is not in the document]
manuelpolzhofer: Yes, no strong opinion of adding it.

manuelpolzhofer (approved):

PR #35, opened 2021-10-12
Add give

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/51, closes https://github.com/radicle-dev/radicle-streaming/issues/30.

Adds `give` method, which gives funds to another user ready for immediate collection. Names "send" and "transfer" are already taken in Solidity :weary: 

~~There's no way for a sub-sender to `give`, is that going to be a problem? Sub-senders are needed only for maintaining stream configurations, so it doesn't make sense for a sub-sender to `give`. On the other hand maybe it's important for bookkeeping? Now that I'm thinking about it, do we need `give` on the funding pool level @manuelpolzhofer ? It's needed for the funding NFT contract, will it use that method for anything else than giving funds from itself to itself? OTOH it allows dripping to others :thinking:~~

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -85,6 +85,27 @@ contract ERC20Pool is Pool {
             );
     }
 
+    /// @notice Gives funds from the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param receiver The receiver
+    /// @param amt The sent amount
+    function give(address receiver, uint128 amt) public {
+        _giveInternal(msg.sender, receiver, amt);
+    }
+
+    /// @notice Gives funds from the sub-sender of the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param subSenderId The ID of the sub-sender
+    /// @param receiver The receiver
+    /// @param amt The given amount
+    function giveFromSubSender(
+        uint256 subSenderId,
+        address receiver,
+        uint128 amt
+    ) public {
+        _giveFromSubSenderInternal(msg.sender, subSenderId, receiver, amt);
+    }
+
     function _transfer(address userAddr, int128 amt) internal override {
         if (amt > 0) erc20.transfer(userAddr, uint128(amt));
         else if (amt < 0) erc20.transferFrom(userAddr, address(this), uint128(-amt));
--- a/src/EthPool.sol
+++ b/src/EthPool.sol
@@ -77,6 +77,21 @@ contract EthPool is Pool {
             );
     }
 
+    /// @notice Gives funds from the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param receiver The receiver
+    function give(address receiver) public payable {
+        _giveInternal(msg.sender, receiver, uint128(msg.value));
+    }
+
+    /// @notice Gives funds from the sub-sender of the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param subSenderId The id of the giver's sub-sender
+    /// @param receiver The receiver
+    function giveFromSubSender(uint256 subSenderId, address receiver) public payable {
+        _giveFromSubSenderInternal(msg.sender, subSenderId, receiver, uint128(msg.value));
+    }
+
     function _transfer(address userAddr, int128 amt) internal override {
         // Take into account the amount already transferred into the pool
         amt += int128(uint128(msg.value));
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -137,6 +137,24 @@ abstract contract Pool {
     /// @param amt The dripped amount
     event Dripped(address indexed sender, address indexed receiver, uint128 amt);
 
+    /// @notice Emitted when funds are given to the receiver.
+    /// @param giver The address of the giver
+    /// @param receiver The receiver
+    /// @param amt The sent amount
+    event Given(address indexed giver, address indexed receiver, uint128 amt);
+
+    /// @notice Emitted when funds are given from the sub-sender to the receiver.
+    /// @param giver The address of the giver
+    /// @param subSenderId The ID of the giver's sub-sender
+    /// @param receiver The receiver
+    /// @param amt The sent amount
+    event GivenFromSubSender(
+        address indexed giver,
+        uint256 indexed subSenderId,
+        address indexed receiver,
+        uint128 amt
+    );
+
     struct Sender {
         // Timestamp at which the funding period has started
         uint64 startTime;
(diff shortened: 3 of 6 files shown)


src/EthPool.sol
```diff
@@ -89,6 +89,13 @@ contract EthPool is Pool {
             );
     }
 
+    /// @notice Gives funds from the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param receiver The receiver
+    function give(address receiver) public payable {
+        _giveInternal(msg.sender, receiver, uint128(msg.value));
```
manuelpolzhofer, line 96: I think this line is missing?
`_transferToContract(msg.sender, uint128(msg.value));`

src/Pool.sol
```diff
@@ -322,6 +351,39 @@ abstract contract Pool {
         }
     }
 
+    function _collectCycles(address receiverAddr, uint64 count)
```
manuelpolzhofer, line 354: I think this is already outdated and we removed the `lastFundsPerCycle`

manuelpolzhofer (commented):
I remember we discussed it in the chat. 

I think it is not possible to `give` to a a subSender but it should be possible that a subsender is the `giver`.

src/EthPool.sol
CodeSandwich: Outdated

src/Pool.sol
CodeSandwich: Outdated

CodeSandwich:
Added `giveFromSubSender`. It acts exactly the same as `give`, but it adds the sub-sender ID to the emitted event :shrug:

src/Pool.sol
```diff
@@ -345,6 +363,50 @@ abstract contract Pool {
...
+    /// @param giverAddr The address of the giver
+    /// @param receiverAddr The receiver
+    /// @param amt The given amount
+    function _giveInternal(
+        address giverAddr,
+        address receiverAddr,
+        uint128 amt
+    ) internal {
+        emit Given(giverAddr, receiverAddr, amt);
+        _giveFromAnyGiver(giverAddr, receiverAddr, amt);
```
manuelpolzhofer, line 377: I would switch the order. In case a of a failed transaction (missing approval) it will be a bit cheaper if the emit is after the potential revert.

src/Pool.sol
```diff
@@ -345,6 +363,50 @@ abstract contract Pool {
         receiver.nextCollectedCycle = cycle;
     }
 
+    /// @notice Gives funds to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param giverAddr The address of the giver
+    /// @param receiverAddr The receiver
+    /// @param amt The given amount
+    function _giveInternal(
+        address giverAddr,
```
manuelpolzhofer, line 372: The `giverAddr` always needs to be msg.sender otherwise it is a huge security risk. I could just use the address of another pool user to give his funds. 

I think we can remove it and just use `msg.sender`

manuelpolzhofer (commented):
True, the only difference is the event. However, currently we have seperate events for sender and subsender 🤔

src/ERC20Pool.sol
```diff
@@ -85,6 +85,27 @@ contract ERC20Pool is Pool {
...
+        uint256 subSenderId,
+        address receiver,
+        uint128 amt
+    ) public {
+        _giveFromSubSenderInternal(msg.sender, subSenderId, receiver, amt);
+    }
+
     function _transfer(address userAddr, int128 amt) internal override {
         if (amt > 0) erc20.transfer(userAddr, uint128(amt));
         else if (amt < 0) erc20.transferFrom(userAddr, address(this), uint128(-amt));
```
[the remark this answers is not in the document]
CodeSandwich: `amt == 0` should be a noop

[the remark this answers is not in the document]
manuelpolzhofer: Ah, missed that. 👍

src/Pool.sol
CodeSandwich: I wouldn't care too much about revert gas cost, it's a rare case when a transaction is submitted despite being reverting.

What I'm trying to achieve here is putting `_transfer` as the very last thing the contract does, because it's a call to an external contract. In this particular function it probably doesn't matter much, but generally it's a good habit, which can prevent nasty scenarios like a re-entry attack.

src/Pool.sol
CodeSandwich: `giverAddr` being `msg.sender` is enforced by `EthPool` and `ERC20Pool`, `_giveInternal` is an internal tool and nobody can call it freely.

This function follows the pattern of `Pool` never touching `msg.sender`. For example `_updateSenderInternal` too accepts the sender address and leaves decision of who it is to the higher-level implementations. We've introduced that pattern back in the `NFTPool` days and since then (AFAIR) we've had a consensus that it's a clean separation of concerns.

src/Pool.sol
manuelpolzhofer: > I wouldn't care too much about revert gas cost, it's a rare case when a transaction is submitted despite being reverting.
It is not so rare, especially if gas prices are high.  You might have pending transaction and you resubmit a new one. Metamask won't revert because approval is here both not for both, etc.

But not a lot of savings, I agree. 

I agree with the re-entrance attack. Normally the rule is bookkeeping first afterwards the payout. 


In our case we transfer the DAI to the contract so maybe in that case the bookkeeping after we received the DAI.

src/Pool.sol
manuelpolzhofer: True. We basically assume correct usage on the high level. It is anyway the same for updateSender as well. So i guess it is fine.

manuelpolzhofer: I think the name `_giveFromAnyGiver` started this thought train. Because you can only give from msg.sender but I guess you indentation was any address can give but obviously only from their address :)

CodeSandwich: It's an unfortunate attempt to follow `_updateSenderInternal` - `_updateSubSenderInternal` - `_updateAnySender` :grimacing:

src/EthPool.sol
```diff
@@ -77,6 +77,21 @@ contract EthPool is Pool {
...
+
+    /// @notice Gives funds from the sub-sender of the sender of the message to the receiver.
+    /// The receiver can collect them immediately.
+    /// @param subSenderId The id of the giver's sub-sender
+    /// @param receiver The receiver
+    function giveFromSubSender(uint256 subSenderId, address receiver) public payable {
+        _giveFromSubSenderInternal(msg.sender, subSenderId, receiver, uint128(msg.value));
+    }
+
     function _transfer(address userAddr, int128 amt) internal override {
```
manuelpolzhofer, line 95: In the give case of an ETH pool. we are actually done and the checks and sending back are not required.

We could make a method `giveTransfer` which for the erc20 case just calls `transfer` and is empty for the ETH pool but I guess its more logic and not really worth it.

manuelpolzhofer (approved):
looks good

src/EthPool.sol
CodeSandwich: Yeah, it's stupid that the ETH pool does that step now. But it simplifies logic at a cost of two arithmetic operations.

PR #36, opened 2021-10-14
Add flushCycles

CodeSandwich:
Depends on drips: https://github.com/radicle-dev/radicle-streaming/pull/33

Adds an escape hatch in case of too many collectable cycles to collect in a single transaction. Hopefully never used.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -267,6 +267,40 @@ abstract contract Pool {
         emit Collected(receiverAddr, collected, dripped);
     }
 
+    /// @notice Counts cycles which will need to be analyzed when collecting or flushing.
+    /// This function can be used to detect that there are too many cycles
+    /// to analyze in a single transaction and flushing is needed.
+    /// @param receiverAddr The address of the receiver
+    /// @return flushable The number of cycles which can be flushed
+    function flushableCycles(address receiverAddr) public view returns (uint64 flushable) {
+        uint64 nextCollectedCycle = receivers[receiverAddr].nextCollectedCycle;
+        if (nextCollectedCycle == 0) return 0;
+        uint64 currFinishedCycle = _currTimestamp() / cycleSecs;
+        return currFinishedCycle + 1 - nextCollectedCycle;
+    }
+
+    /// @notice Flushes uncollected cycles of the user.
+    /// Flushed cycles won't need to be analyzed when the user collects from them.
+    /// Calling this function does not collect and does not affect the collectable amount.
+    ///
+    /// This function is needed when collecting funds received over a period so long, that the gas
+    /// needed for analyzing all the uncollected cycles can't fit in a single transaction.
+    /// Calling this function allows spreading the analysis cost over multiple transactions.
+    /// A cycle is never flushed more than once, even if this function is called many times.
+    /// @param receiverAddr The address of the receiver
+    /// @param maxCycles The maximum number of flushed cycles.
+    /// If too low, flushing will be cheap, but will cut little gas from the next collection.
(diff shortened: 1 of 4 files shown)


manuelpolzhofer (commented):
nice that we have that now 🚀

src/Pool.sol
```diff
@@ -267,6 +267,54 @@ abstract contract Pool {
...
+        uint128 collected = 0;
+        int128 fundsPerCycle = int128(receiver.lastFundsPerCycle);
+        uint64 collectedCycle = receiver.nextCollectedCycle;
+        for (uint64 i = 0; i < flushed; i++) {
+            fundsPerCycle += receiver.amtDeltas[collectedCycle - 1].nextCycle;
+            delete receiver.amtDeltas[collectedCycle - 1];
+            fundsPerCycle += receiver.amtDeltas[collectedCycle].thisCycle;
+            collected += uint128(fundsPerCycle);
+            collectedCycle++;
+        }
```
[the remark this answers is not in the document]
CodeSandwich: Added a helper `_collectCycles`, which extracts the common logic.

IMO we should take a look and deduplicate even more logic around collecting, but that's probably a good material for a separate PR.

src/Pool.sol
```diff
@@ -267,6 +267,40 @@ abstract contract Pool {
...
+    /// A cycle is never flushed more than once, even if this function is called many times.
+    /// @param receiverAddr The address of the receiver
+    /// @param maxCycles The maximum number of flushed cycles.
+    /// If too low, flushing will be cheap, but will cut little gas from the next collection.
+    /// If too high, flushing may become too expensive to fit in a single transaction.
+    /// @return flushable The number of cycles which can be flushed
+    function flushCycles(address receiverAddr, uint64 maxCycles) public returns (uint64 flushable) {
+        Receiver storage receiver = receivers[receiverAddr];
+        flushable = flushableCycles(receiverAddr);
+        uint64 flushed = maxCycles < flushable ? maxCycles : flushable;
```
manuelpolzhofer, line 298:
```suggestion
        uint64 cycles = maxCycles < flushable ? maxCycles : flushable;
```
The past tense is a bit confusing because the `flush` didn't happen yet.

src/Pool.sol
```diff
@@ -283,19 +317,8 @@ abstract contract Pool {
...
-                lastFundsPerCycle += receiver.amtDeltas[collectedCycle - 1].nextCycle;
-                delete receiver.amtDeltas[collectedCycle - 1];
-                lastFundsPerCycle += receiver.amtDeltas[collectedCycle].thisCycle;
-                collected += uint128(lastFundsPerCycle);
-            }
-            receiver.lastFundsPerCycle = uint128(lastFundsPerCycle);
-            receiver.nextCollectedCycle = collectedCycle;
-        }
+        uint64 cycles = flushableCycles(receiverAddr);
+        collected += _collectCycles(receiverAddr, cycles);
```
manuelpolzhofer, line 321: 👍

src/Pool.sol
```diff
@@ -322,6 +345,25 @@ abstract contract Pool {
         }
     }
 
+    function _collectCycles(address receiverAddr, uint64 count)
```
manuelpolzhofer, line 348: I think we should agree what a collect means

-collect
  - flushCycle
  - drip

If thats the case, we should call this method flushCycleInternal :)

manuelpolzhofer (approved):
LGTM.

some ideas for better naming

src/Pool.sol

src/Pool.sol
CodeSandwich: Good point! Fixed

PR #38, opened 2021-10-16
Cheap collect

CodeSandwich:
Based on realization that `amtDeltas[cycleX].nextCycle` will never be updated when `cycleX` is finished. That means that when `cycleX` is collected, value of its `nextCycle` is final and can be safely consumed and deleted.

When a receiver has `amtDelta` set for the first time ever, let's say that for `cycleY`, `amtDeltas[cycleY].thisCycle` is set, never `amtDeltas[cycleY-1].nextCycle`, so the latter doesn't need to be ever analyzed.

This PR reduces two SLOADs to one for each collected cycle.


manuelpolzhofer:
the past tense for the variable `nextCollectedCycle` is a bit confusing. maybe we could rename it. As an alternative we could also just store the `lastCollected` cycle.

PR #42, opened 2021-10-20
Drop storage of lastFundsPerCycle

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-streaming/issues/39. Saves ~250K gas when running all tests.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -179,9 +179,6 @@ abstract contract Pool {
         // The next cycle to be collected
         uint64 nextCollectedCycle;
         // --- SLOT BOUNDARY
-        // The amount of funds received for the last collected cycle.
-        uint128 lastFundsPerCycle;
-        // --- SLOT BOUNDARY
         // The changes of collected amounts on specific cycle.
         // The keys are cycles, each cycle `C` becomes collectable on timestamp `C * cycleSecs`.
         mapping(uint64 => AmtDelta) amtDeltas;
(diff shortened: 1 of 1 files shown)


src/Pool.sol
```diff
@@ -351,16 +348,16 @@ abstract contract Pool {
...
-            collectedAmt += uint128(fundsPerCycle);
-            fundsPerCycle += receiver.amtDeltas[cycle].nextCycle;
+            cycleAmt += receiver.amtDeltas[cycle].thisCycle;
+            collectedAmt += uint128(cycleAmt);
+            cycleAmt += receiver.amtDeltas[cycle].nextCycle;
             delete receiver.amtDeltas[cycle];
             cycle++;
         }
-        receiver.lastFundsPerCycle = uint128(fundsPerCycle);
+        if (cycleAmt != 0) receiver.amtDeltas[cycle].thisCycle += cycleAmt;
```
manuelpolzhofer, line 360: 💯

manuelpolzhofer: Maybe we should add a comment somewhere. 
If it is the next cycle to collect the delta mapping doesn't store the delta instead it includes the cycleAmt

CodeSandwich: I've just tried it in `_flushCyclesInternal` and the whole tests started using 1048 gas more. I don't understand why. It'd be a good price for cleaner code, but IMO it's basically the same.

PR #44, opened 2021-10-22
Drop storage of weights list properties

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/41. Drops storage of metadata about the receivers list, which can be cheaply extracted from calldata.


Julia77th:
We can discuss it

PR #45, opened 2021-10-26
Add custom errors

byterose:
Custom errors decrease both deploy and runtime gas costs. 
Additionally all expected errors are listed in the abi file which can be used to improve the UX of contract interactions.

[Introduction blog post](https://blog.soliditylang.org/2021/04/21/custom-errors/)

- [x] Add custom errors
- [ ] Update tests

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -151,6 +151,23 @@ abstract contract Pool {
     /// @param amt The dripped amount
     event Dripped(address indexed sender, address indexed receiver, uint128 amt);
 
+    /// @notice Invalid total dripped amount.
+    error InvalidTotalDrippedAmount();
+
+    /// The amount `amt` is higher than available funds of sender.
+    /// @param amt The amount to be withdrawn, must not be higher than available funds.
+    error NotEnoughFundsInSenderAccount(uint128 amt);
+
+    /// Drip fraction `dripsFraction` is too high.
+    /// @param dripsFraction The fraction of received funds to be dripped.
+    error InvalidDripFraction(uint32 dripsFraction);
+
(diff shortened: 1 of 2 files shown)


gh0stwheel:
@byterose -- thank you for your contribution!

(CC @manuelpolzhofer)

@CodeSandwich -- a few days ago, I think you asked me to take a look at this and share my input. I wasn't familiar with these new errors either, so I did some quick reading about them. From my quick research, I don't believe that this is something that's worth switching over to at this stage of the project, given that our main goal right now is to get the MVP deployed as quickly as possible, and given that we're not sure how these errors would interact with wallets like Metamask. It's also not clear to me that there would be major gas savings, and that gas savings would only exist in corner cases in any event.

I would propose that we add this to a backlog of tasks to work on after the MVP has been deployed. At that time, if we wanted to evaluate this, I would propose that we write a few test cases to evaluate gas savings and also test how it works with Metamask.

Thoughts?

byterose:
@gh0stwheel thank you for your input.

One thing to keep in mind is that a `require` statement uses a revert underneath. If you use a `require` statement with a string argument, it will revert with that string as error data. To see how this would look like at a Yul level, have a look at my previous comment. In conclusion not much would change, since both use revert underneath.

I also recommend to watch this [video](https://www.youtube.com/watch?v=fWuMpfPuGRM), custom errors are introduced at **16:50**.


P.S. I did not find a way yet to use `DSTest` to test reverts, that's why the CI is failing as well.

CodeSandwich:
Let's shelve it for now. I've opened an issue #61 to keep track of that idea.

PR #47, opened 2021-10-29
Switch receivers configuration from weights to amtPerSec

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/44.

This is a major shift in the pool API, drops the notion of weights in favor of plain amtPerSecs. This PR is divided into 2 commits and it's probably the best approach to view them one after another:
1. updates the logic and the tests in the most concise way
2. renames things and fixes comments without altering the logic

The gas usage of all calls to `updateSender` in all tests drops from 6,970,631 to 5,670,301 (19%).

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.8.7;
 
-struct ReceiverWeight {
+struct Receiver {
     address receiver;
     uint128 amtPerSec;
 }
@@ -11,18 +11,15 @@ struct ReceiverWeight {
 /// The contract has 2 types of users: the senders and the receivers.
 ///
 /// A sender has some funds and a set of addresses of receivers, to whom he wants to send funds.
-/// In order to send there are 3 conditions, which must be fulfilled:
+/// In order to send there are 2 conditions, which must be fulfilled:
 ///
 /// 1. There must be funds on his account in this contract.
 ///    They can be added with `topUp` and removed with `withdraw`.
-/// 2. Total amount sent to the receivers every second must be set to a non-zero value.
-///    This is done with `setAmtPerSec`.
-/// 3. A set of receivers must be non-empty.
+/// 2. A set of receivers must be non-empty.
 ///    Receivers can be added, removed and updated with `setReceiver`.
-///    Each receiver has a weight, which is used to calculate how the total sent amount is split.
 ///
 /// Each of these functions can be called in any order and at any time, they have immediate effects.
-/// When all of these conditions are fulfilled, every second the configured amount is being sent.
+/// When both conditions are fulfilled, every second the configured amount is being sent.
 /// It's extracted from the `withdraw`able balance and transferred to the receivers.
 /// The process continues automatically until the sender's balance is empty.
 ///
@@ -30,7 +27,7 @@ struct ReceiverWeight {
 /// A sub-sender is identified by a user address and an ID.
 /// The sender and all sub-senders' configurations are independent and they have separate balances.
 ///
-/// A receiver has an account, from which he can `collect` funds sent by the senders.
+/// A receiver has an account, from which they can `collect` funds sent by the senders.
 /// The available amount is updated every `cycleSecs` seconds,
 /// so recently sent funds may not be `collect`able immediately.
 /// `cycleSecs` is a constant configured when the pool is deployed.
@@ -58,9 +55,9 @@ abstract contract Pool {
     uint64 internal constant MAX_TIMESTAMP = type(uint64).max - 2;
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
-    uint32 public constant SENDER_WEIGHTS_COUNT_MAX = 100;
+    uint32 public constant MAX_RECEIVERS = 100;
     /// @notice Maximum value of drips fraction
-    uint32 public constant DRIPS_FRACTION_MAX = 1_000_000;
+    uint32 public constant MAX_DRIPS_FRACTION = 1_000_000;
     /// @notice The amount passed as the withdraw amount to withdraw all the funds
     uint128 public constant WITHDRAW_ALL = type(uint128).max;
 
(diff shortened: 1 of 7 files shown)


src/Pool.sol
```diff
@@ -755,50 +657,46 @@ abstract contract Pool {
...
+    /// @param receivers The list of the user's receivers.
     /// @param updates The list of stream updates to log
     function _startSending(
         Sender storage sender,
-        ReceiverWeight[] calldata weights,
+        Receiver[] calldata receivers,
         StreamUpdates memory updates
     ) internal {
-        uint32 weightsSum = _weightsSum(weights);
+        uint128 amtPerSec = _totalAmtPerSec(receivers);
```
manuelpolzhofer, line 674: I think it would be cheaper to let the user provide the `_totalAmtPerSec` as a parameter. 

A iteration is happening in the `_setDeltasFromNow`. Here we can calculate the acutal totalAmtPerSec which could be returned. 

If the provided `totalAmtPerSec` doesn't match with the actual one we `revert`.

In that way we force the user to provide the correct one

src/Pool.sol
```diff
@@ -327,21 +312,18 @@ abstract contract Pool {
...
         // Dripped when collected
         if (collected > 0 && currReceivers.length > 0 && sender.dripsFraction > 0) {
-            uint256 drippable = (uint256(collected) * sender.dripsFraction) / DRIPS_FRACTION_MAX;
-            uint32 weightsSum = _weightsSum(currReceivers);
-            dripped = uint128(drippable - (drippable % weightsSum));
-            collected -= dripped;
-            uint128 drippedPerWeight = dripped / weightsSum;
-            uint128 actuallyDripped = 0;
+            uint256 drippable = (uint256(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION;
+            uint128 totalAmtPerSec = _totalAmtPerSec(currReceivers);
```
manuelpolzhofer, line 316: Same here: `totalAmtPerSec` could be provided as parameter and afterwards calc the acutal one and compare both.

src/Pool.sol
```diff
@@ -587,140 +533,98 @@ abstract contract Pool {
...
+    /// where 0 means no dripping and `MAX_DRIPS_FRACTION` dripping everything.
     function getDripsFraction(address userAddr) public view returns (uint32 dripsFraction) {
         return senders[userAddr].dripsFraction;
     }
 
-    function _assertCurrReceivers(Sender storage sender, ReceiverWeight[] calldata currReceivers)
+    /// @notice Asserts that the list of receivers is the sender's currently used one.
+    /// @param sender The sender
+    /// @param currReceivers The list of the user's current receivers.
+    function _assertCurrReceivers(Sender storage sender, Receiver[] calldata currReceivers)
```
manuelpolzhofer, line 558: Why not public?

src/Pool.sol
```diff
@@ -587,140 +533,98 @@ abstract contract Pool {
...
-    /// @param newReceivers The list of the user's receivers and their weights,
-    /// which shall be in use after this function is called.
-    function _setReceivers(Sender storage sender, ReceiverWeight[] calldata newReceivers) internal {
-        require(newReceivers.length <= SENDER_WEIGHTS_COUNT_MAX, "Too many receivers");
-        uint256 senderWeightSum = 0;
-        for (uint256 i = 0; i < newReceivers.length; i++) {
-            senderWeightSum += newReceivers[i].weight;
-            require(newReceivers[i].weight > 0, "Receiver weight zero");
+    /// @param receivers The new list of the user's receivers
+    function _setReceivers(Sender storage sender, Receiver[] calldata receivers) internal {
```
manuelpolzhofer, line 613: maybe

```suggestion
    function _setReceiversHash(Sender storage sender, Receiver[] calldata receivers) internal {
```

src/Pool.sol
```diff
@@ -587,140 +533,98 @@ abstract contract Pool {
...
+            totalAmtPerSec += receivers[i].amtPerSec;
             if (i > 0) {
-                address prevReceiver = newReceivers[i - 1].receiver;
-                address currReceiver = newReceivers[i].receiver;
+                address prevReceiver = receivers[i - 1].receiver;
+                address currReceiver = receivers[i].receiver;
                 require(prevReceiver <= currReceiver, "Receivers not sorted by address");
                 require(prevReceiver != currReceiver, "Duplicate receivers");
             }
         }
```
manuelpolzhofer, line 625: I think this might be cheaper, if we reduce the lines of code in the loop
```solidity
        require(receivers[0].receiver != 0, "Receiver amtPerSec is zero");
        uint256 totalAmtPerSec = receivers[0].receiver.amtPerSec;
        for (uint256 i = 1; i < receivers.length; i++) {
            require(receivers[i].amtPerSec != 0, "Receiver amtPerSec is zero");
            totalAmtPerSec += receivers[i].amtPerSec;
            address prevReceiver = receivers[i - 1].receiver;
            address currReceiver = receivers[i].receiver;
            require(prevReceiver <= currReceiver, "Receivers not sorted by address");
            require(prevReceiver != currReceiver, "Duplicate receivers");
        }
```

src/Pool.sol
```diff
@@ -755,50 +657,46 @@ abstract contract Pool {
...
-        uint128 amtPerWeight = sender.amtPerSec / weightsSum;
-        uint128 amtPerSec = amtPerWeight * weightsSum;
+        if (amtPerSec == 0) return;
         // Won't be sending anything
         if (sender.startBalance < amtPerSec) return;
         sender.startTime = _currTimestamp();
         uint256 endTimeUncapped = _currTimestamp() + uint256(sender.startBalance / amtPerSec);
         uint64 endTime = endTimeUncapped > MAX_TIMESTAMP ? MAX_TIMESTAMP : uint64(endTimeUncapped);
-        _setDeltasFromNow(weights, int128(amtPerWeight), endTime, updates);
+        _setDeltasFromNow(receivers, 1, endTime, updates);
```
manuelpolzhofer, line 682: Maybe we can make a constant for the negative and positive multiplier and pass it here?

manuelpolzhofer (approved):
🚀 looks good overall.

Some small comments, regarding potential gas savings. Some we discussed already in the previous PR.

src/Pool.sol
CodeSandwich: The multiplier is a quick and dirty solution, the whole `_setDeltasFromNow` will be removed in the next PR

src/Pool.sol
CodeSandwich: Because it accepts `Sender storage sender`

src/Pool.sol
CodeSandwich: We've discussed it here: https://github.com/radicle-dev/radicle-streaming/pull/44#discussion_r736018052, https://github.com/radicle-dev/radicle-streaming/pull/44#discussion_r736020501, https://github.com/radicle-dev/radicle-streaming/pull/44#discussion_r73602260, https://github.com/radicle-dev/radicle-streaming/pull/41#discussion_r736001457 and we've opened an issue for that: https://github.com/radicle-dev/radicle-streaming/issues/46

src/Pool.sol
manuelpolzhofer: Ah, i see

src/Pool.sol
manuelpolzhofer: okay. perfecto

src/Pool.sol
```diff
@@ -56,18 +53,13 @@ abstract contract Pool {
     uint64 public immutable cycleSecs;
     /// @dev Timestamp at which all funding periods must be finished
     uint64 internal constant MAX_TIMESTAMP = type(uint64).max - 2;
-    /// @notice Maximum sum of all receiver weights of a single sender.
-    /// Limits loss of per-second funding accuracy, they are always multiples of weights sum.
-    uint32 public constant SENDER_WEIGHTS_SUM_MAX = 10_000;
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
-    uint32 public constant SENDER_WEIGHTS_COUNT_MAX = 100;
+    uint32 public constant MAX_RECEIVERS = 100;
```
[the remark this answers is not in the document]
CodeSandwich: It's arbitrary, we can make it larger. 100 seems big enough to be useful while being safe when it comes to maximum gas usage in the context of the maximum block size.

src/Pool.sol

src/Pool.sol
CodeSandwich: We'd cut out a single `if` on a local variable in a loop, I really don't think that it's worth the code duplication

PR #48, opened 2021-11-04
Update all sender's streams in a single loop

CodeSandwich:


src/Pool.sol
```diff
@@ -453,14 +453,113 @@ abstract contract Pool {
...
+        uint64 currEndTime,
+        uint64 newEndTime
+    ) internal returns (StreamUpdates memory updates) {
+        // Skip iterating over `currReceivers` if funding has run out
+        uint256 currIdx = currEndTime > _currTimestamp() ? 0 : currReceivers.length;
+        // Skip iterating over `newReceivers` if no new funding is started
+        uint256 newIdx = newEndTime > _currTimestamp() ? 0 : newReceivers.length;
+        uint256 updatesMax = currReceivers.length + newReceivers.length - currIdx - newIdx;
+        updates = StreamUpdates({length: 0, updates: new StreamUpdate[](updatesMax)});
+        while (true) {
```
manuelpolzhofer, line 510: This can be a bit simplified. Tests are passing.
```solidity
        while (currIdx < currReceivers.length || newIdx < newReceivers.length) {
            // Each iteration gets the next stream update and applies it on the receiver.
            // A stream update is composed of two receiver configurations, one from each list,
            // or from a single configuration if the receiver is only on one of the lists.
            bool pickCurr = currIdx < currReceivers.length;
            bool pickNew = newIdx < newReceivers.length;
            if (pickCurr && pickNew) {
                pickCurr = currReceivers[currIdx].receiver <= newReceivers[newIdx].receiver;
                pickNew = newReceivers[newIdx].receiver <= currReceivers[currIdx].receiver;
            }
            address receiver = address(0);
            int128 currAmt = 0;
            int128 newAmt = 0;
            if (pickCurr) {
                receiver = currReceivers[currIdx].receiver;
                currAmt = int128(currReceivers[currIdx].amtPerSec);
                _setDelta(receiver, currEndTime, currAmt);
                currIdx++;
            }
            if (pickNew) {
                receiver = newReceivers[newIdx].receiver;
                newAmt = int128(newReceivers[newIdx].amtPerSec);
                _setDelta(receiver, newEndTime, -newAmt);
                newIdx++;
            }
            _setDelta(receiver, _currTimestamp(), newAmt - currAmt);
            updates.updates[updates.length++] = StreamUpdate({
            receiver: receiver,
            amtPerSec: uint128(newAmt),
            endTime: newAmt == 0 ? _currTimestamp() : newEndTime
            });
            // The receiver was never used, initialize it.
            if (!pickCurr && receiverStates[receiver].nextCollectedCycle == 0) {
                receiverStates[receiver].nextCollectedCycle = _currTimestamp() / cycleSecs + 1;
            }
        }
```

manuelpolzhofer (approved):
very nice optimisation. The algorithm seems correct. I still want to look again at the tests. 
Maybe we can also try to define an invariant for the tests.

src/Pool.sol
CodeSandwich: I've made a diff and the only change I can see here is that we repeatedly check relationship between the index and the list length instead of doing it once and reusing the result, is it simpler?

src/Pool.sol
```diff
@@ -453,14 +453,113 @@ abstract contract Pool {
...
+    /// @return withdrawn The withdrawn amount which should be sent to the user.
+    /// Equal to `withdrawAmt` unless `WITHDRAW_ALL` is used.
+    function _updateSenderBalance(
+        Sender storage sender,
+        uint128 topUpAmt,
+        uint128 withdrawAmt,
+        uint128 amtPerSec,
+        uint64 endTime
+    ) internal returns (uint128 withdrawn) {
+        if (endTime > _currTimestamp()) endTime = _currTimestamp();
```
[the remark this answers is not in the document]
CodeSandwich: Good catch, bug getting the block timestamp is a single opcode costing 2 gas, not much room for improvement :laughing:

src/Pool.sol
```diff
@@ -453,14 +453,113 @@ abstract contract Pool {
...
+    /// @param sender The analyzed sender
+    /// @param amtPerSec The sender's total amount per second
+    /// @return sendingEndTime The sending end time
+    function _sendingEndTime(Sender storage sender, uint128 amtPerSec)
+        internal
+        view
+        returns (uint64 sendingEndTime)
+    {
+        if (amtPerSec == 0) return _currTimestamp();
+        uint256 endTime = sender.startTime + uint256(sender.startBalance / amtPerSec);
```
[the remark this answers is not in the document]

src/Pool.sol
manuelpolzhofer: I think it is cleaner to remove the `if else`  and move the condition when the loops stop into the `while(...)`

manuelpolzhofer: Maybe we can also add the cases to the comments. From my notes: 

```javascript
// cases: 
// case 1 only currReceiver left: pick only currReceiver
// case 2 only nextReceiver left: pick only nextReceiver
// case 3 currReceiver and nextReceiver are left and both are equal: pickCurr and pickNext
// case 4 currReceiver and nextReceiver are left and both are not equal: pick smaller one
```

CodeSandwich: > I think it is cleaner to remove the `if else`  and move the condition when the loops stop into the `while(...)`

The nice property of how it's expressed now is that it's clearly exhaustive, it's easy to see what should happen in case there's nothing to pick. If we remove `if else`, the algorithm is still exhaustive, but only by coincidence, which IMO is both harder to follow and easier to introduce bugs. I don't have very a strong opinion on that, if you're certain that your proposition will make ours and newcomers' lives easier, let's fix that.

manuelpolzhofer: It is already highly optimised the trade-off is it is a bit hard to understand. 
Especially to see all the cases when `pickCurr` and `pickNew` are happening.

I think some more comments would be nice.   

I leave the decision for the `while` up to you. In the end it is taste

CodeSandwich: Added some comments. The problem with this algorithm is that there are really not that many steps in it, but it's not trivial. It connects the `amtDelta`s updates, which is a tough concept on its own, with iteration matching pairs from two sorted lists, which is a rather uncommon way of consuming lists.

manuelpolzhofer (approved):

src/Pool.sol
```diff
@@ -453,14 +453,122 @@ abstract contract Pool {
...
+        uint256 updatesMax = currReceivers.length + newReceivers.length - currIdx - newIdx;
+        updates = StreamUpdates({length: 0, updates: new StreamUpdate[](updatesMax)});
+        while (true) {
+            // Each iteration gets the next stream update and applies it on the receiver.
+            // A stream update is composed of two receiver configurations, one current and one new,
+            // or from a single receiver configuration if the receiver is being added or removed.
+            bool pickCurr = currIdx < currReceivers.length;
+            bool pickNew = newIdx < newReceivers.length;
+            if (!pickCurr && !pickNew) break;
+            else if (pickCurr && pickNew) {
```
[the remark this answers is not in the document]
CodeSandwich: Less noisy, fixed

PR #50, opened 2021-11-04
Emit stream update events in-place

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/49

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -172,15 +172,10 @@ abstract contract Pool {
         int128 nextCycle;
     }
 
-    struct StreamUpdates {
-        uint256 length;
-        StreamUpdate[] updates;
-    }
-
-    struct StreamUpdate {
-        address receiver;
-        uint128 amtPerSec;
-        uint64 endTime;
+    struct SenderId {
+        bool isSubSender;
+        address senderAddr;
+        uint256 subSenderId;
     }
 
     /// @dev Details about all the senders, the key is the owner's address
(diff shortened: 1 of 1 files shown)


src/Pool.sol
```diff
@@ -372,9 +367,10 @@ abstract contract Pool {
     {
         (collected, dripped) = _collectInternal(senderAddr, currReceivers);
         Sender memory sender = senders[senderAddr];
-        StreamUpdates memory updates;
-        (withdrawn, updates) = _updateAnySender(
+        SenderId memory senderId = SenderId(false, senderAddr, 0);
```
manuelpolzhofer, line 370: I think this format has higher readability 
```suggestion
        SenderId memory senderId = SenderId({isSubSender: false,...});
```

src/Pool.sol
```diff
@@ -538,18 +516,40 @@ abstract contract Pool {
...
+    /// @param amtPerSec The new amount per second sent from the sender to the receiver
+    /// or 0 if sending is stopped
+    /// @param endTime The timestamp when the funds stop being sent.
+    function emitStreamUpdated(
+        SenderId memory senderId,
+        address receiver,
+        uint128 amtPerSec,
+        uint64 endTime
+    ) internal {
+        if (amtPerSec == 0) endTime = _currTimestamp();
```
manuelpolzhofer, line 539: I would move this line to the `_updateStreams`. For me the emit function should just dealing with emitting the correct event.

manuelpolzhofer (approved):
looks good. small comments

src/Pool.sol
CodeSandwich: IMO it's the best place for this logic. It simplifies the API of `emitStreamUpdated` and guarantees consistency, because now it's impossible to forget about this special case. It's also logic specific for emitting the event, not for updating streams, so it's a good separation of concerns.

src/Pool.sol
CodeSandwich: Added helper functions, should be even more readable

PR #51, opened 2021-11-04
Transfer ERC-20 only once per transaction

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/50, closes https://github.com/radicle-dev/radicle-streaming/issues/37

### DIFF
--- a/src/EthPool.sol
+++ b/src/EthPool.sol
@@ -77,7 +77,11 @@ contract EthPool is Pool {
             );
     }
 
-    function _transfer(address to, uint128 amt) internal override {
-        if (amt != 0) payable(to).transfer(amt);
+    function _transfer(address userAddr, int128 amt) internal override {
+        // Take into account the amount already transferred into the pool
+        amt += int128(uint128(msg.value));
+        if (amt == 0) return;
+        require(amt > 0, "Sending a negative ether amount");
+        payable(userAddr).transfer(uint128(amt));
     }
 }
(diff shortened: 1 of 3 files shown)


src/EthPool.sol
```diff
@@ -77,7 +77,11 @@ contract EthPool is Pool {
             );
     }
 
-    function _transfer(address to, uint128 amt) internal override {
-        if (amt != 0) payable(to).transfer(amt);
+    function _transfer(address userAddr, int128 amt) internal override {
+        // Take into account the amount already transferred into the pool
+        amt += int128(uint128(msg.value));
+        if (amt == 0) return;
+        require(amt > 0, "Sending a negative ether amount");
```
manuelpolzhofer, line 84:
```suggestion
        require(amt > 0, "not-enough-eth-sent");
```

manuelpolzhofer: Might be easier to understand for the user why it failed

manuelpolzhofer: eth transfer logic seems to work

manuelpolzhofer (approved):
looks good

src/EthPool.sol
CodeSandwich: This is more of a sanity check and a message to the developer who breaks the transfer logic. It should never fail, the user can't break it, but can't fix it either, sending more eth won't help :D

PR #52, opened 2021-11-10
Split drips without reminder

CodeSandwich:
A minor tweak to the drips splitting algorithm making it split cleanly, without the reminder. Previously the reminder was always sent to the receiver, even if dripping was set to 100%. This was wasting gas and potentially causing issues with contracts not expecting receiving transfers.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -240,13 +240,9 @@ abstract contract Pool {
         }
 
         // Dripped when collected
-        Sender storage sender = senders[receiverAddr];
-        if (collected > 0 && currReceivers.length > 0 && sender.dripsFraction > 0) {
-            uint256 drippable = (uint256(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION;
-            uint128 totalAmtPerSec = _totalAmtPerSec(currReceivers);
-            for (uint256 i = 0; i < currReceivers.length; i++) {
-                dripped += uint128((drippable * currReceivers[i].amtPerSec) / totalAmtPerSec);
-            }
+        if (collected > 0 && currReceivers.length > 0) {
+            Sender storage sender = senders[receiverAddr];
+            dripped = uint128((uint160(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION);
             collected -= dripped;
         }
     }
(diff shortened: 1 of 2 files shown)


src/Pool.sol
```diff
@@ -240,13 +240,9 @@ abstract contract Pool {
...
-        Sender storage sender = senders[receiverAddr];
-        if (collected > 0 && currReceivers.length > 0 && sender.dripsFraction > 0) {
-            uint256 drippable = (uint256(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION;
-            uint128 totalAmtPerSec = _totalAmtPerSec(currReceivers);
-            for (uint256 i = 0; i < currReceivers.length; i++) {
-                dripped += uint128((drippable * currReceivers[i].amtPerSec) / totalAmtPerSec);
-            }
+        if (collected > 0 && currReceivers.length > 0) {
+            Sender storage sender = senders[receiverAddr];
+            dripped = uint128((uint160(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION);
```
manuelpolzhofer, line 245: 👍  True, the iteration is not required because we don't want to modify the receivers in the collectable

src/Pool.sol
```diff
@@ -325,10 +321,11 @@ abstract contract Pool {
...
             uint256 drippable = (uint256(collected) * sender.dripsFraction) / MAX_DRIPS_FRACTION;
             uint128 totalAmtPerSec = _totalAmtPerSec(currReceivers);
+            uint128 drippedAmtPerSec = 0;
             for (uint256 i = 0; i < currReceivers.length; i++) {
-                uint128 dripAmt = uint128(
-                    (drippable * currReceivers[i].amtPerSec) / totalAmtPerSec
-                );
+                drippedAmtPerSec += currReceivers[i].amtPerSec;
+                uint128 dripAmt = uint128((drippable * drippedAmtPerSec) / totalAmtPerSec) -
+                    dripped;
```
manuelpolzhofer, line 328: Why? I see that this approach is also working but is it not just more complicated?

CodeSandwich: Let's assume that we have receivers `R1` with `amtPerSec ==  1` and `R2` with `amtPerSec == 1`. We've configured `dripsFraction` to be 100%. We've received 101 wei and we're calling `collect`. The total 101 is `drippable`, `totalAmtPerSec == 2`.

According to the old algorithm `R1` gets `101 * 1 / 2 = 50` and `R2` gets `101 * 1 / 2 = 50`. `dripped == 100`, `collected == drippable - dripped == 1`. We're getting a transfer of 1 despite having `dripsFraction` 100%.

With the new algorithm `R1` gets `101 * 1 / 2 - 0 = 50`, but `R2` gets `101 * 2 / 2 = 51`. We're getting a transfer of 0, everything has been dripped.

manuelpolzhofer: Ah, I see. It solves the problem but it is still a bit arbitrary  that `R2` would benefit. We could also say the project should keep the rest. It would be a bit complicated to calculate it in advance. I guess it is still necessary to do the iteration. We could just compare the `dripped` in the end.   

No strong opinion. The amounts are so tiny anyway.

CodeSandwich: Yeah, the amounts are tiny and drips can't be split with perfect justice, because `totalAmtPerSec` ususally won't be a multiple of `drippable`. IMO some random injustice in the range of +/- 1 wei is fine to have.

You're right about calculation in advance. Another issue is that if we keep the reminder uncollected hoping that the next cycle will make it available, the whole API stops adding up perfectly. What should `collectable` return? The value actually collectable and hide that some received funds or the total value? If the actual value, the pool introduces schroedinger's tokens, which sometimes disappear and reappear. If the total value, `collectable` stops describing `collect`.

manuelpolzhofer: True, that would be also a bit weird. I guess then the current approach is the best.

manuelpolzhofer (approved):

PR #53, opened 2021-11-10
Add setDripsReceiver

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/52. This is the first PR for making the drips configuration independent from the sender's receivers list. It allows storing the drips receivers, but they aren't used anywhere yet.

### DIFF
--- a/src/EthPool.sol
+++ b/src/EthPool.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.8.7;
 
-import {Pool, Receiver} from "./Pool.sol";
+import {DripsReceiver, Pool, Receiver} from "./Pool.sol";
 
 /// @notice Funding pool contract for Ether.
 /// See the base `Pool` contract docs for more details.
@@ -90,6 +90,20 @@ contract EthPool is Pool {
     /// @param receiver The receiver
     function giveFromSubSender(uint256 subSenderId, address receiver) public payable {
         _giveFromSubSenderInternal(msg.sender, subSenderId, receiver, uint128(msg.value));
+    }
+
+    /// @notice Sets a new list of drips receivers of the sender of the message.
+    /// @param currReceivers The list of the user's drips receivers which is currently in use.
+    /// If this function is called for the first time for the user, should be an empty array.
+    /// @param newReceivers The new list of the user's drips receivers.
+    /// Must be sorted by the drips receivers' addresses, deduplicated and without 0 weights.
+    /// Each drips receiver will be getting `weight / TOTAL_DRIPS_WEIGHTS`
+    /// share of the funds collected by the user.
+    function setDripsReceivers(
+        DripsReceiver[] calldata currReceivers,
+        DripsReceiver[] calldata newReceivers
+    ) public {
+        _setDripsReceiversInternal(msg.sender, currReceivers, newReceivers);
     }
 
     function _transfer(address userAddr, int128 amt) internal override {
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -6,6 +6,11 @@ struct Receiver {
     uint128 amtPerSec;
 }
 
+struct DripsReceiver {
+    address receiver;
+    uint32 weight;
+}
+
 /// @notice Funding pool contract. Automatically sends funds to a configurable set of receivers.
 ///
 /// The contract has 2 types of users: the senders and the receivers.
@@ -56,6 +61,11 @@ abstract contract Pool {
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant MAX_RECEIVERS = 100;
+    /// @notice Maximum number of drips receivers of a single user.
+    /// Limits costs of dripping.
+    uint32 public constant MAX_DRIPS_RECEIVERS = 200;
+    /// @notice The total drips weights of a user
+    uint32 public constant TOTAL_DRIPS_WEIGHTS = 1_000_000;
     /// @notice Maximum value of drips fraction
     uint32 public constant MAX_DRIPS_FRACTION = 1_000_000;
     /// @notice The amount passed as the withdraw amount to withdraw all the funds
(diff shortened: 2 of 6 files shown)


src/EthPool.sol
```diff
@@ -92,6 +92,20 @@ contract EthPool is Pool {
...
     }
 
+    /// @notice Sets a new list of drips receivers of the sender of the message.
+    /// @param currReceivers The list of the user's drips receivers which is currently in use.
+    /// If this function is called for the first time for the user, should be an empty array.
+    /// @param newReceivers The new list of the user's drips receivers.
+    /// Must be sorted by the drips receivers' addresses, deduplicated and without 0 weights.
+    /// Each drips receiver will be getting `weight / TOTAL_DRIPS_WEIGHTS`
+    /// share of the funds collected by the user.
+    function setDripsReceivers(
```
manuelpolzhofer, line 102: Why is this method on the ETHPool and ERC20 level? Both are doing the same could we not move it to the pool?

src/Pool.sol
```diff
@@ -56,6 +61,11 @@ abstract contract Pool {
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant MAX_RECEIVERS = 100;
+    /// @notice Maximum number of drip sreceivers of a single user.
+    /// Limits costs of dripping.
+    uint32 public constant MAX_DRIPS_RECEIVERS = 200;
```
manuelpolzhofer, line 66: :1

src/Pool.sol
```diff
@@ -56,6 +61,11 @@ abstract contract Pool {
     /// @notice Maximum number of receivers of a single sender.
     /// Limits costs of changes in sender's configuration.
     uint32 public constant MAX_RECEIVERS = 100;
+    /// @notice Maximum number of drip sreceivers of a single user.
```
manuelpolzhofer, line 64:
```suggestion
    /// @notice Maximum number of drips receivers of a single user.
```

src/Pool.sol
```diff
@@ -773,6 +791,75 @@ abstract contract Pool {
...
+    /// @param newReceivers The new list of the user's drips receivers.
+    /// Must be sorted by the drips receivers' addresses, deduplicated and without 0 weights.
+    /// Each drips receiver will be getting `weight / TOTAL_DRIPS_WEIGHTS`
+    /// share of the funds collected by the user.
+    function _setDripsReceiversInternal(
+        address userAddr,
+        DripsReceiver[] calldata currReceivers,
+        DripsReceiver[] calldata newReceivers
+    ) internal {
+        _assertCurrDripsReceivers(userAddr, currReceivers);
```
manuelpolzhofer, line 807: We don't care about the currReceiver in my opinion. We can just overwrite it. 

However, should we not check that the `collect` is up to date. I guess that will be in the other PRs :)

src/Pool.sol
```diff
@@ -773,6 +791,75 @@ abstract contract Pool {
...
+        uint64 totalWeight = 0;
+        for (uint256 i = 0; i < receivers.length; i++) {
+            require(receivers[i].weight != 0, "Drips receiver weight is zero");
+            totalWeight += receivers[i].weight;
+            if (i > 0) {
+                address prevReceiver = receivers[i - 1].receiver;
+                address currReceiver = receivers[i].receiver;
+                require(prevReceiver <= currReceiver, "Drips receivers not sorted by address");
+                require(prevReceiver != currReceiver, "Duplicate drips receivers");
+            }
```
manuelpolzhofer, line 827: Why are potential duplicates a problem? I can't see any major downside.

src/EthPool.sol
CodeSandwich: We could, but then `Pool` would be dealing with `msg.sender`. We want to keep that separation and let the non-abstract pools decide which user is being updated.

src/Pool.sol
CodeSandwich: It'll be important in the next PR :) Here it just enforces the correct usage of the API, which doesn't do any meaningful work yet.

src/Pool.sol
CodeSandwich: That's a good point, I haven't thought about it! Unlike the sender receivers the drips receivers don't have to be consumed in a structured fashion. They could be zeroed, unsorted and duplicated, they'll still work. The only downside would be that it would generate unnecessary gas cost to have unnecessary entries and that it would emit noisy events. If we want to ban duplicates, we need to enforce sorting too, these requirements are bundle. If we don't enforce being sorted and stripped of zeroed entries, it may be also confusing that two equivalent lists of receivers can't be used as substitutes, because they generate different hashes.

What do you think? Should we keep the API strict or should we loosen it?

src/Pool.sol
CodeSandwich: Fixed

src/Pool.sol
CodeSandwich: I hope it's a `:+1:`, not `:-1:` :laughing:

manuelpolzhofer: haha

src/Pool.sol
```diff
@@ -196,6 +211,10 @@ abstract contract Pool {
         uint256 subSenderId;
     }
 
+    /// @notice Current drips configuration hash, see `hashDripsReceivers`.
+    /// The key is the user address.
+    mapping(address => bytes32) public dripsReceiversHash;
```
manuelpolzhofer, line 216: Why here and not in sender struct? I think it is fine as well, just curious.

src/Pool.sol
manuelpolzhofer: In general i agree with strictness but in this case it costs gas.
However, we anyway need to iterate the list to check if `totalWeight` is smaller equal than `TOTAL_DRIPS_WEIGHTS`.

This we obviously can't skip.

manuelpolzhofer: Therefore, we can keep it to ensure the order.

manuelpolzhofer (approved):
🚀

PR #54, opened 2021-11-10
Use drips receivers

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-streaming/pull/53. Switches drips from using the senders' receivers lists to drips receivers lists. It's probably the best to review commit by commit, each of them is self-contained.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -744,7 +744,7 @@ abstract contract Pool {
         return subSenders[senderAddr][subSenderId].receiversHash;
     }
 
-    /// @notice Sets a new list of drips receivers of a user
+    /// @notice Collects received funds and sets a new list of drips receivers of the user.
     /// @param userAddr The user address
     /// @param currReceivers The list of the user's drips receivers which is currently in use.
     /// If this function is called for the first time for the user, should be an empty array.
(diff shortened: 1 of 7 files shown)


src/Pool.sol
```diff
@@ -336,21 +323,21 @@ abstract contract Pool {
...
-            uint128 drippedAmtPerSec = 0;
+        if (collected > 0 && currReceivers.length > 0) {
+            uint32 drippedWeight = 0;
             for (uint256 i = 0; i < currReceivers.length; i++) {
-                drippedAmtPerSec += currReceivers[i].amtPerSec;
-                uint128 dripAmt = uint128((drippable * drippedAmtPerSec) / totalAmtPerSec) -
-                    dripped;
+                drippedWeight += currReceivers[i].weight;
+                uint128 dripAmt = uint128(
+                    (uint160(collected) * drippedWeight) / TOTAL_DRIPS_WEIGHTS - dripped
```
manuelpolzhofer, line 331: I like it. It will also simplify the usage. I want to give 1% of my funds to project X and 5% to project Y.

Instead of first defining a dripFraction and after that the percentages. 👍

src/Pool.sol
```diff
@@ -792,23 +744,26 @@ abstract contract Pool {
...
+    /// @return collected The collected amount
+    /// @return dripped The amount dripped to the user's receivers
     function _setDripsReceiversInternal(
         address userAddr,
         DripsReceiver[] calldata currReceivers,
         DripsReceiver[] calldata newReceivers
-    ) internal {
-        _assertCurrDripsReceivers(userAddr, currReceivers);
+    ) internal returns (uint128 collected, uint128 dripped) {
+        (collected, dripped) = _collectInternal(userAddr, currReceivers);
```
manuelpolzhofer, line 762: I am wondering about the following scenario.

Lets assume i set my drips to `project A` at the beginning of the cycle. A lot of projects start supporting me because I am also very generous in dripping. However, just before the cycle ends I could change my drips again. Correct?

Because in `_collectInternal` the cycle is still ongoing. 

Only funds I would have received from other drips would be impacted.

If thats the case, I am wondering if we should allow to only change your drips once in a cycle. Obviously we would have to store the cycle/block.timestamp or put it in the hash.

manuelpolzhofer (approved):
👍

src/Pool.sol
CodeSandwich: That's a valid concern, which we've discussed when we were designing drips. Your scenario can be further generalized: we can look at the last drips update OR the beginning of the current cycle, whichever is later. In both cases there's a period of time, when people have been streaming to you and thought that you'd drip it to somebody different than in reality.

This generalization makes the proposed solution too weak. The attacker doesn't need to make two updates in a single cycle to redirect an almost full cycle of streams to unwanted drips.

A solution could be to make the new configuration active only since the next cycle, this way there would be no way to change the drips destination of already streamed, but not yet collectable funds. This however would make it difficult to react to malicious receivers of said drips, they would be guaranteed to keep receiving drips at least until the end of the current cycle. I'm not sure if that would improve the ecosystem, but it would certainly make the pool much more complex, both in the API and in its implementation. We'd probably need to have two drips configurations, the old one for the current cycle and a new one for the future ones.

If you have an idea, how to solve the current cycle drips problem, don't hesitate to propose it, it would be great to have a safer ecosystem!

PR #57, opened 2021-11-16
Use SenderId in internal functions

CodeSandwich:
This allows removing some duplicate logic and clean up functions naming. It's a purely refactoring PR.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -145,14 +145,14 @@ abstract contract Pool {
     /// @notice Emitted when funds are given to the receiver.
     /// @param giver The address of the giver
     /// @param receiver The receiver
-    /// @param amt The sent amount
+    /// @param amt The given amount
     event Given(address indexed giver, address indexed receiver, uint128 amt);
 
     /// @notice Emitted when funds are given from the sub-sender to the receiver.
     /// @param giver The address of the giver
     /// @param subSenderId The ID of the giver's sub-sender
     /// @param receiver The receiver
-    /// @param amt The sent amount
+    /// @param amt The given amount
     event GivenFromSubSender(
         address indexed giver,
         uint256 indexed subSenderId,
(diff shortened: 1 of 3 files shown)


src/Pool.sol
```diff
@@ -475,22 +397,55 @@ abstract contract Pool {
...
         _updateStreams(senderId, currReceivers, newReceivers, currEndTime, newEndTime);
+
+        _storeSender(senderId, sender);
+        _emitSenderUpdated(senderId, sender.startBalance, newReceivers);
+        _transfer(senderId.senderAddr, int128(withdrawn) - int128(topUpAmt));
+    }
+
+    function _loadSender(SenderId memory senderId) internal view returns (Sender memory) {
+        if (senderId.isSubSender) {
+            return subSenders[senderId.senderAddr][senderId.subSenderId];
```
manuelpolzhofer, line 425: Did we compare the gas difference if we would store both in one global mapping based on the hash of `senderAddr` and `subSenderId`?

One pattern for upgradeability is to implement some kind of storage contract with getter and setter methods. 

Maybe we can discuss this a bit, seems like we heading a bit anyway towards this direction with `load` and `store` methods.

On the other hand it will increase the gas costs again, so i guess it is fine.

manuelpolzhofer (approved):
looks good.

src/Pool.sol
CodeSandwich: The gas usage would increase a little. In the current setup a `sender` is stored at slot
```solidity
keccak256(
  concat(
    uint256(senderId.senderAddr),
    SENDER_MAPPING_SLOT
  )
)
```
while a sub-sender at slot
```solidity
keccak256(
  concat(
    senderId.subSenderId,
    keccak256(
      concat(
        uint256(senderId.senderAddr),
        SUB_SENDER_MAPPING_SLOT
      )
   )
)
```
IIUC what you're proposing would store a sender at:
```solidity
keccak256(
  concat(
    keccak256(
      concat(
        senderId.senderAddr,
        senderId.isSubSender,
        senderId.isSubSender ? sender.subSenderId : 0
      ),
      SUB_SENDER_MAPPING_SLOT
   )
)
```
For sub-senders it's pretty much the same thing, but for senders it hashes one time more.

To put it in context, hashing costs 30 gas + 6 gas per hashed word and we're accessing storage no more than twice per call. Obfuscating data structure is not worth any savings in this area.

PR #58, opened 2021-11-16
Move the whole sender state off-chain

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-streaming/issues/46.

### DIFF
--- a/src/Pool.sol
+++ b/src/Pool.sol
@@ -108,15 +108,15 @@ abstract contract Pool {
 
     /// @notice Emitted when a sender is updated
     /// @param sender The updated sender
-    /// @param balance The sender's balance since the event block's timestamp
-    /// @param receivers The list of the user's receivers.
+    /// @param balance The new sender's balance
+    /// @param receivers The new list of the sender's receivers.
     event SenderUpdated(address indexed sender, uint128 balance, Receiver[] receivers);
 
-    /// @notice Emitted when a sender is updated
+    /// @notice Emitted when a sub-sender is updated
     /// @param senderAddr The address of the sender
     /// @param subSenderId The id of the sender's updated sub-sender
-    /// @param balance The sender's balance since the event block's timestamp
-    /// @param receivers The list of the user's receivers.
+    /// @param balance The sub-sender's balance
+    /// @param receivers The new list of the sub-sender's receivers.
     event SubSenderUpdated(
         address indexed senderAddr,
         uint256 indexed subSenderId,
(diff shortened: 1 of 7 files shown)


src/Pool.sol
```diff
@@ -563,110 +585,141 @@ abstract contract Pool {
...
+        SenderId memory senderId,
+        uint128 newBalance,
+        Receiver[] calldata newReceivers
+    ) internal {
+        bytes32 stateHash = hashSenderState(_currTimestamp(), newBalance, newReceivers);
+        if (senderId.isSubSender) {
+            subSenderStateHash[senderId.senderAddr][senderId.subSenderId] = stateHash;
+        } else {
+            senderStateHash[senderId.senderAddr] = stateHash;
+        }
```
manuelpolzhofer, line 703: I think we should have a seperate `UpdateSenderHash`event. The `_currTimestamp`/ `lastUpdate` information is currently not part of any event.  Implict because available because of the block.timestamp.

src/Pool.sol
```diff
@@ -563,110 +585,141 @@ abstract contract Pool {
...
+    /// @param currReceivers The list of receivers set in the last update of the sender.
     /// @return withdrawableAmt The withdrawable amount
-    function _withdrawable(SenderId memory senderId, Receiver[] calldata currReceivers)
-        internal
-        view
-        returns (uint128 withdrawableAmt)
-    {
-        Sender memory sender = _loadSender(senderId);
-        _assertCurrReceivers(sender, currReceivers);
+    function _withdrawable(
```
manuelpolzhofer, line 649: I am currently questioning a bit the sense of the `view` methods like `_withdrawable`.   Before the PR the contract stored the state and performed a calculation based on the state.  Which makes sense.

Now, we need to provide the state to perform the calculation.
Then the question is why the method exits at all, I could perform the witthdrawable calcuation off-chain in a libary and maybe only get the hash from the contract to verify.

On the other hand for other applications on-chain it might be useful.  

One idea would be to make the  formulars in pure methods and move it to a library.

src/Pool.sol
manuelpolzhofer: However, it might be nice to have the data and the hash in the same event.

CodeSandwich: `block.timestamp` of emission is a part of EVERY event :smile: `SenderUpdated` event contains all the data needed to both recreate the hash and call `updateSender`, IMO putting hash into the event would be just a waste of gas, if one needs the hash off-chain, they can easily calculate it for free.

src/Pool.sol
CodeSandwich: It's a good observation and I totally agree! You'll like https://github.com/radicle-dev/radicle-streaming/pull/60 :)

src/Pool.sol
manuelpolzhofer: Yeah, I know it is available. I think if we provide a low-level javascript library everything should be fine.

manuelpolzhofer (approved):
nice work 🤖

PR #64, opened 2021-11-23
Rename sub-sender to account

CodeSandwich:
Unfortunately `alias` is a solidity keyword :grimacing: IMO `account` is fine too.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -21,19 +21,19 @@ struct DripsReceiver {
 /// Every second funds are deducted from the sender balance and sent to their receivers.
 /// The process stops automatically when the sender's balance is not enough to cover another second.
 ///
-/// A single address can act as any number of independent senders by using sub-senders.
-/// A sub-sender is identified by a user address and an ID.
-/// The sender and all sub-senders' configurations are independent and they have separate balances.
+/// A single address can act as any number of independent senders by using accounts.
+/// An account is identified by a user address and an account identifier.
+/// The sender and their accounts' configurations are independent and they have separate balances.
 ///
-/// A receiver has an account, from which they can `collect` funds sent by the senders.
+/// A receiver has a balance, from which they can `collect` funds sent by the senders.
 /// The available amount is updated every `cycleSecs` seconds,
 /// so recently sent funds may not be `collect`able immediately.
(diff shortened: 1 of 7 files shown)


src/DripsHub.sol
```diff
@@ -172,10 +172,10 @@ abstract contract DripsHub {
         int128 nextCycle;
     }
 
-    struct SenderId {
-        bool isSubSender;
-        address senderAddr;
-        uint256 subSenderId;
+    struct UserOrAccount {
```
manuelpolzhofer, line 175: I would prefer a generic term here.

Sender
  - User
  - Account
  
Maybe Sender or something else which can be either a user or an account?

src/DripsHub.sol
```diff
@@ -103,14 +103,14 @@ abstract contract DripsHub {
...
-    /// @notice Emitted when a sub-sender is updated
+    /// @notice Emitted when a sender account is updated
     /// @param senderAddr The address of the sender
-    /// @param subSenderId The id of the sender's updated sub-sender
-    /// @param balance The sub-sender's balance
-    /// @param receivers The new list of the sub-sender's receivers.
+    /// @param account The sender's account
+    /// @param balance The account's balance
+    /// @param receivers The new list of the account's receivers.
     event SenderUpdated(
```
manuelpolzhofer, line 111: We still call the event Sender here so i guess it is fine?

manuelpolzhofer (approved):
I like the term account.

src/DripsHub.sol
CodeSandwich: We'll be dropping the notion of a `sending` in favor of dripping, so `sender` will go too. E.g. `event SenderUpdated` will be `event DripsUpdated`.

manuelpolzhofer: okay

src/DripsHub.sol
CodeSandwich: It's going to be hard to get any more generic :D `UserOrAccount` is clunky, but it doesn't limit its usage and clearly reuses existing terms. `Sender` would reintroduce a new term (we're dropping it) and it would make `_give(Sender memory sender, ...)` signature a bit weird.

PR #70, opened 2021-11-26
Optimize _totalDripsAmtPerSec

CodeSandwich:
Depends on https://github.com/radicle-dev/radicle-drips-hub/pull/69. Cuts gas of `_totalDripsAmtPerSec` by 60% or by ~17K for the maximum 100 receivers.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -461,6 +461,25 @@ abstract contract DripsHub {
         return uint128(amtPerSec);
     }
 
+    /// @notice Calculates the total amount per second of all the drips receivers.
+    /// @param receivers The list of the receivers.
+    /// It must have passed `_assertDripsReceiversValid` in the past.
+    /// @return totalAmtPerSec The total amount per second of all the drips receivers
+    function _totalDripsAmtPerSec(DripsReceiver[] memory receivers)
+        internal
+        pure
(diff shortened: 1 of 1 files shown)


src/DripsHub.sol
```diff
@@ -461,6 +461,25 @@ abstract contract DripsHub {
...
+    /// @return totalAmtPerSec The total amount per second of all the drips receivers
+    function _totalDripsAmtPerSec(DripsReceiver[] memory receivers)
+        internal
+        pure
+        returns (uint128 totalAmtPerSec)
+    {
+        uint256 length = receivers.length;
+        uint256 i = 0;
+        while (i < length) {
+            // Safe, because `receivers` passed `_assertDripsReceiversValid` in the past
```
manuelpolzhofer, line 476:
```suggestion
            // Safe, because `receivers` passed `_assertDripsReceiversValid` in the past and  there is not enough gas for `i` to produce an overflow 
```

manuelpolzhofer (approved):
nice optimisation 🚀

src/DripsHub.sol
CodeSandwich: `_assertDripsReceiversValid` guarantees that `receivers.length <= MAX_DRIPS_RECEIVERS`, so `i <= 100`, no gas  considerations are needed

Issue #56, opened 2021-11-12
Add a treasury


CodeSandwich:
The funds currently locked in the pool aren't used for anything. We may be able to automatically invest them if it's safe, e.g. in [DSR](https://makerdao.world/en/learn/Dai/dsr/). The first step in this experiment is to add a simple treasury, which as of now won't do anything except keeping the funds.

PR #72, opened 2021-12-02
Add ERC-20 reserve

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/56.

It may make sense to require that the user of the hub approves spending of ERC-20 not by the hub itself, but by `hub.reserve()`. It'd save us an unnecessary transfer, but it could be weirder for the users of the hub.

### DIFF
--- a/src/DaiReserve.sol
+++ b/src/DaiReserve.sol
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.7;
+
+import {ERC20Reserve, IERC20Reserve} from "./ERC20Reserve.sol";
+import {IDai} from "./Dai.sol";
+
+interface IDaiReserve is IERC20Reserve {
+    function dai() external view returns (IDai);
+}
+
+contract DaiReserve is ERC20Reserve, IDaiReserve {
+    IDai public immutable override dai;
+
+    constructor(
+        IDai _dai,
+        address owner,
+        address user
+    ) ERC20Reserve(_dai, owner, user) {
+        dai = _dai;
+    }
+}
--- a/src/ERC20DripsHub.sol
+++ b/src/ERC20DripsHub.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.7;
 
 import {SplitsReceiver, DripsHub, DripsReceiver} from "./DripsHub.sol";
+import {IERC20Reserve} from "./ERC20Reserve.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
 
 /// @notice Drips hub contract for any ERC-20 token.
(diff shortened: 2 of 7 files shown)


src/ERC20Reserve.sol
```diff
@@ -0,0 +1,60 @@
...
+        require(erc20.transfer(_msgSender(), amt), "Reserve: transfer failed");
+    }
+
+    function deposit(uint256 amt) public override onlyUser {
+        balance += amt;
+        emit Deposited(_msgSender(), amt);
+        require(erc20.transferFrom(_msgSender(), address(this), amt), "Reserve: transfer failed");
+    }
+
+    function forceWithdraw(uint256 amt) public onlyOwner {
```
manuelpolzhofer, line 51: I am wondering if we should just have one `withdraw` method which can be used by both `user` and `owner`. Since the user is a smart contract which doesn't have the calls for `setUser` implemented, we could also just use a `mapping` with one auth type.

No, strong opinion however. I think owner should be able to take the funds without first assigning the user role to itself.

src/ERC20DripsHub.sol
```diff
@@ -17,12 +23,15 @@ contract ERC20DripsHub is DripsHub {
     /// @param owner The initial owner of the contract with managerial abilities.
     /// @param _erc20 The address of an ERC-20 contract which tokens the drips hub will work with.
     /// The supply of the tokens must be lower than `2 ^ 127`.
+    /// @param _reserve The address of the funds reserve to be used.
     constructor(
         uint64 cycleSecs,
         address owner,
-        IERC20 _erc20
+        IERC20 _erc20,
```
manuelpolzhofer, line 30: I wouldn't pass the erc20 here, it is just a way for additional potential misstakes. 

We could get it from the reserve. `reserve.erc20()` or add the check if they are the same. 

I think it makes sense to store the address of the erc20 in the hub contract to make the approval  calls cheaper

src/ERC20Reserve.sol
manuelpolzhofer: Ahh, now i saw the difference the `forceWithdraw` allows you to take all of it. (if we have more the balance)

src/ERC20Reserve.sol
```diff
@@ -0,0 +1,60 @@
...
+        transferOwnership(owner);
+        setUser(_user);
+    }
+
+    modifier onlyUser() {
+        require(_msgSender() == user, "Reserve: caller is not the user");
+        _;
+    }
+
+    function withdraw(uint256 amt) public override onlyUser {
```
manuelpolzhofer, line 38: The `withdraw` would work with `withdraw(address usr, uint256 amt)` but it would break the consistency. :)

manuelpolzhofer: Without going over the hub

manuelpolzhofer (approved):
looks good overall.

Some small ideas and comments but no blockers. 👍

src/ERC20DripsHub.sol
CodeSandwich: Clever! Fixed

src/ERC20Reserve.sol
CodeSandwich: Yeah, I'm not a fan of any of these solutions either

src/DaiReserve.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity ^0.8.7;
+
+import {ERC20Reserve, IERC20Reserve} from "./ERC20Reserve.sol";
+import {IDai} from "./Dai.sol";
+
+interface IDaiReserve is IERC20Reserve {
+    function dai() external view returns (IDai);
+}
+
+contract DaiReserve is ERC20Reserve, IDaiReserve {
```
manuelpolzhofer, line 11: Why is it important to have a dedicated Dai reserve?

src/ERC20DripsHub.sol
```diff
@@ -135,6 +135,7 @@ contract ERC20DripsHub is DripsHub {
     /// @notice Set the new reserve address.
     /// @param newReserve The new reserve.
     function setReserve(IERC20Reserve newReserve) public onlyOwner {
+        require(newReserve.erc20() == erc20, "Invalid reserve ERC-20 address");
```
manuelpolzhofer, line 138: 👍

src/DaiReserve.sol
CodeSandwich: Mostly for clarity and type safety, so we don't need to blindly cast `IERC20` to `IDai`. It's not very important, but it's not a lot of complex code either.

manuelpolzhofer (approved):

PR #73, opened 2021-12-07
Add DripsHub proxy

CodeSandwich:
It puts all DripsHubs behind a proxy making them upgradable using the UUPS pattern.

Moves the owner and paused state to the unstructured storage. To do that the PR adds new mixins. The next PR #74 moves the rest of the DripsHub storage to unstructured storage. This way any upgrades in any part of the inheritance tree can be done atomically and without affecting the storage layout.

### DIFF
--- a/src/ManagedDripsHub.sol
+++ b/src/ManagedDripsHub.sol
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.7;
+
+import {UUPSUpgradeable} from "openzeppelin-contracts/proxy/utils/UUPSUpgradeable.sol";
+import {ERC1967Proxy} from "openzeppelin-contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import {ERC1967Upgrade} from "openzeppelin-contracts/proxy/ERC1967/ERC1967Upgrade.sol";
+import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
+import {DripsHub, SplitsReceiver} from "./DripsHub.sol";
+
+/// @notice The DripsHub which is UUPS-upgradable, pausable and has an admin.
+/// It can't be used directly, only via a proxy.
+///
+/// ManagedDripsHub uses the ERC-1967 admin slot to store the admin address.
+/// All instances of the contracts are owned by address `0x00`.
+/// While this contract is capable of updating the admin,
+/// the proxy is expected to set up the initial value of the ERC-1967 admin.
+///
+/// All instances of the contracts are paused and can't be unpaused.
+/// When a proxy uses such contract via delegation, it's initially unpaused.
+abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable {
+    /// @notice The ERC-1967 storage slot for the contract.
+    /// It holds a single boolean indicating if the contract is paused.
+    bytes32 private constant SLOT_PAUSED =
+        bytes32(uint256(keccak256("eip1967.managedDripsHub.paused")) - 1);
+
+    /// @notice Emitted when the pause is triggered.
+    /// @param account The account which triggered the change.
+    event Paused(address account);
+
+    /// @notice Emitted when the pause is lifted.
+    /// @param account The account which triggered the change.
+    event Unpaused(address account);
+
+    /// @notice Initializes the contract in paused state and with no admin.
+    /// The contract instance can be used only as a call delegation target for a proxy.
+    /// @param cycleSecs The length of cycleSecs to be used in the contract instance.
+    /// Low value makes funds more available by shortening the average time of funds being frozen
+    /// between being taken from the users' drips balances and being collectable by their receivers.
+    /// High value makes collecting cheaper by making it process less cycles for a given time range.
+    constructor(uint64 cycleSecs) DripsHub(cycleSecs) {
+        _pausedSlot().value = true;
+    }
+
+    /// @notice Collects all received funds available for the user
+    /// and transfers them out of the drips hub contract to that user's wallet.
+    /// @param user The user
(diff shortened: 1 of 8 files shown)


src/ERC1967Ownable.sol
```diff
@@ -0,0 +1,35 @@
...
+import {ERC1967Upgrade} from "openzeppelin-contracts/proxy/ERC1967/ERC1967Upgrade.sol";
+
+/// @notice The proxy-safe ownability mix-in.
+/// It's supposed to be used only in contracts behind a proxy.
+///
+/// This contract uses the ERC-1967 admin slot to store the owner address.
+/// While this contract is capable of updating the owner,
+/// the proxy is expected to set up the initial value of the ERC-1967 admin.
+/// All instances of `ERC1967Ownable` contracts have owner set and locked to `0x00`.
+abstract contract ERC1967Ownable is ERC1967Upgrade {
```
manuelpolzhofer, line 13: Open Zeppelin doesn't offer an upgradable ownership contract, i assume?

src/ManagedDripsHub.sol
```diff
@@ -0,0 +1,75 @@
...
+import {DripsHub, SplitsReceiver} from "./DripsHub.sol";
+import {ERC1967Pausable} from "./ERC1967Pausable.sol";
+
+/// @notice The DripsHub which is UUPS-upgradable, pausable and has an owner.
+/// It can't be used directly, only via a proxy.
+///
+/// ManagedDripsHub uses the ERC-1967 admin slot to store the owner address.
+/// While this contract is capable of updating the owner,
+/// the proxy is expected to set up the initial value of the ERC-1967 admin.
+abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable, ERC1967Pausable {
```
manuelpolzhofer, line 17: I think the `MangedDripsHub` is a bit over engineered. We could remove it and just make
```solidity
ERC20DripHub is DripsHub, UUPSUpgradeable, ERC1967Pausable

```

What do you think?

src/ERC1967Ownable.sol
CodeSandwich: They don't, they only have simple ERC-1967 `admin` functions, which are rather bare bones. The classic `Ownable` isn't suitable here, even if gutted, because it uses a storage slot and you can't stop it from doing so, because it writes to it in the constructor. It also has `renounceOwnership`, which doesn't make sense here.

src/ManagedDripsHub.sol
```diff
@@ -0,0 +1,75 @@
...
+        override
+        whenNotPaused
+        returns (uint64 flushable)
+    {
+        return super.flushCycles(user, maxCycles);
+    }
+
+    /// @notice Authorizes UUPSUpgradable upgrades
+    function _authorizeUpgrade(address newImplementation) internal view override onlyOwner {
+        newImplementation;
```
manuelpolzhofer, line 64: Can you share some additional docs here? In case you have any :)

Why is it not required to call
`_upgradeTo(newImplementation)` here?

manuelpolzhofer: Ah good it. This is just to check the auth pattern

src/ManagedDripsHub.sol
CodeSandwich: We'd need to do that with `EthDripsHub` too. IMO it's the right way to organize code, because it encapsulates some important bits, which don't really match the `ERC20` part of `ERC20DripsHub` and could get obscured in the token logic. After all the goal isn't to have as few files as possible.

src/ManagedDripsHub.sol
manuelpolzhofer: So I would call proxy.upgradeTo(newAddr) and it calls the _authorizedUpgrade to check the owner.

CodeSandwich: Yes. Maybe the docs should say instead:
```solidity
/// @notice See `UUPSUpgradable` docs for more details
```

src/ERC1967Pausable.sol
```diff
@@ -0,0 +1,69 @@
...
+import {ERC1967Ownable} from "./ERC1967Ownable.sol";
+
+/// @notice The proxy-safe pausability mix-in.
+/// Only the owner can pause or unpause the contract.
+/// It's supposed to be used only in contracts behind a proxy.
+///
+/// This contract uses a custom ERC-1967 slot to store its state.
+/// All instances of `ERC1967Pausable` contracts are paused and can't be unpaused.
+/// When a proxy uses such contract via delegation, it's initially unpaused.
+abstract contract ERC1967Pausable is ERC1967Ownable {
```
manuelpolzhofer, line 14: What about removing `ERC1967Ownable` entirely. 

Make the methods virutal.
```

    /// @notice Triggers stopped state.
    function pause() public virtual whenNotPaused  {
        _setPaused(true);
        emit Paused(msg.sender);
    }

    /// @notice Returns to normal state.
    function unpause() public virtual whenPaused  {
        _setPaused(false);
        emit Unpaused(msg.sender);
    }
```

Than bring everything together in the ManagerDrip layer?

manuelpolzhofer (approved):
good work 👍

src/ManagedDripsHub.sol
manuelpolzhofer: Not relevant anymore with new design. Resolved.

src/ERC1967Pausable.sol
manuelpolzhofer: Implemented in the #75

CodeSandwich: Fixed

src/ManagedDripsHub.sol
CodeSandwich: Fixed

PR #74, opened 2021-12-07
Switch DripsHub to unstructured storage

CodeSandwich:
Depends on #73.

### DIFF
--- a/src/ERC20DripsHub.sol
+++ b/src/ERC20DripsHub.sol
@@ -5,14 +5,17 @@ import {SplitsReceiver, DripsReceiver} from "./DripsHub.sol";
 import {ManagedDripsHub} from "./ManagedDripsHub.sol";
 import {IERC20Reserve} from "./ERC20Reserve.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
+import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 
 /// @notice Drips hub contract for any ERC-20 token. Must be used via a proxy.
 /// See the base `DripsHub` and `ManagedDripsHub` contract docs for more details.
 contract ERC20DripsHub is ManagedDripsHub {
+    /// @notice The ERC-1967 storage slot for the contract.
+    /// It holds a single address of the ERC-20 reserve.
+    bytes32 private constant SLOT_RESERVE =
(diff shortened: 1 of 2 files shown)


src/ERC20DripsHub.sol
```diff
@@ -5,14 +5,16 @@ import {SplitsReceiver, DripsReceiver} from "./DripsHub.sol";
...
 import {IERC20Reserve} from "./ERC20Reserve.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
+import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 
 /// @notice Drips hub contract for any ERC-20 token. Must be used via a proxy.
 /// See the base `DripsHub` and `ManagedDripsHub` contract docs for more details.
 contract ERC20DripsHub is ManagedDripsHub {
+    /// @notice The ERC-1967 storage slot for the contract.
+    /// It holds a single address of the ERC-20 reserve.
+    bytes32 private constant SLOT = bytes32(uint256(keccak256("eip1967.erc20DripsHub")) - 1);
```
manuelpolzhofer, line 15: If I understand correctly:

We have the following unstructered slots:
**erc20DripsHub**
- reserveAddress

**dripsHub**
- business logic storage

**ManagerHub**
- no storage

**erc1967Pausable**
- stores pausable

**Ownable**
- no storage

**Upgradablity**
- admin storage (owner)

I am assuming here.
```solidity
contract A {
bytes32 private constant SLOT = "slotA";
}

contract B {
bytes32 private constant SLOT = "slotB";
}

contract C is A, B {
bytes32 private constant SLOT = "slotC";
}
```

That storage operations in contract A use slot A and contract operations in B use slot B.


Is this correct? I am wondering if it is necessary to have seperated slots for owner and pausable?

manuelpolzhofer: Okay, just check it, is correct.

CodeSandwich: Yes, this is correct. The owner uses the standard ERC-1967 slot, which is supposed to hold only the address, nothing else. We could sneak an extra boolean into that slot, but that would be against the standard and could interfere with existing ERC-1967 tooling.

If your concern is gas cost of managing more slots, then don't worry. It's just 1 extra slot set when deploying and it's touched only in `onlyOwner` functions, which are rarely used and don't need to be extremely gas optimized.

manuelpolzhofer (approved):
looks good

PR #75, opened 2021-12-10
feat: move ownable and pause into DripsHubUpgradable

manuelpolzhofer:
Alternative design to simplify: #73


manuelpolzhofer:
Implemented in #73

PR #78, opened 2021-12-15
Add ManagedDripsHub tests

CodeSandwich:


src/test/ManagedDripsHub.t.sol
```diff
@@ -0,0 +1,178 @@
...
+        super.setUp(dripsHub);
+    }
+
+    function createUser() internal override returns (DripsHubUser) {
+        return createManagedUser();
+    }
+
+    function createManagedUser() internal virtual returns (ManagedDripsHubUser);
+
+    function wrapInProxy(ManagedDripsHub hubLogic) internal returns (ManagedDripsHub) {
```
manuelpolzhofer, line 36: clean solution fo the tests :1

src/test/ManagedDripsHub.t.sol
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    function testOnlyAdminCanUpgradeContract() public {
+        try user.upgradeTo(address(0)) {
+            assertTrue(false, "ChangeAdmin hasn't reverted");
+        } catch Error(string memory reason) {
+            assertEq(reason, ERROR_NOT_ADMIN, "Invalid changeAdmin revert reason");
+        }
+        // Fails for admin too, but due to a different reason.
+        // As of now we can't reasonably test contract upgrade.
```
manuelpolzhofer, line 62: Why?

manuelpolzhofer (approved):
nice thats good to have more tests by using the drip hub with a proxy

src/test/ManagedDripsHub.t.sol
CodeSandwich: Because we only have 1 implementation of the pool, so it'd be hard to see any changes. OTOH we could make a "dumb" upgrade, which would change some constants, like cycle length :thinking:

manuelpolzhofer: Exactly, something like that or even the exact same contract but just redeployed

CodeSandwich: Fixed

PR #83, opened 2022-01-11
Add ens support

deckardkrug:
This pr is a proposal to add ens name support.
There are a few things that I would like to get feedback on before I continue.

1. Currently `_getENSAdress` is declared in the `ManagedDripsHub` contract. Should it stay there or might it be better to move it to the `DripsHub` contract?
2. My initial thought is to use function overloading in order to enable passing an ens name hash to every function that would normally accept an address. However this would add a lot of functions that only differentiate by one parameter and a call to `_getENSAddress`. Is there a better way? Maybe we pass addresses as strings and then add a function that evaluates whether the passed string is an address or an ens name hash. Alternatively we could implement an authentication module and use that in front of all relevant functions.

### DIFF
--- a/src/ManagedDripsHub.sol
+++ b/src/ManagedDripsHub.sol
@@ -92,11 +92,11 @@ abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable {
     /// @param hash The ENS name hash.
     /// @return owner Address that the ens name points to.
     function _getENSAddress(string memory hash) internal view returns (address owner) {
-	Resolver resolver = ens.resolver(hash);
-	owner = resolver.addr(hash);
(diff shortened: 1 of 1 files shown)


deckardkrug:
A cool feature we could implement with this is an escrow stream.
The base idea is to enable opening streams towards github users without a radicle account. The reserve would just hold the stream tokens until the user registers his radicle ens name, therefore creating an economic incentive to move to radicle.

Only obstacle would be to check if the user that is registering his `.radicle.eth` ens is the actual maintainer of the project that the stream points to. Maybe the [DNS Registrar](https://docs.ens.domains/contract-api-reference/dns-registrar) is a solution to that obstacle.

deckardkrug:
Closing in favor of our recent discussions.

PR #92, opened 2022-02-14
Make reserve immutable in hub

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-drips-hub/issues/95.

The reserve doesn't need to be changeable on the drips hub side. With the introduction of the multi-token reserve it'll be not only unnecessary but also virtually impossible. The logic without the reserve setter is simpler and safer, also the gas cost for the users is lower because there are fewer storage accesses. This PR doesn't prevent the immutable variable from being changed with a logic upgrade.

CC: [deckardkrug](https://github.com/deckardkrug)

### DIFF
--- a/src/ERC20DripsHub.sol
+++ b/src/ERC20DripsHub.sol
@@ -10,23 +10,24 @@ import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 /// @notice Drips hub contract for any ERC-20 token. Must be used via a proxy.
 /// See the base `DripsHub` and `ManagedDripsHub` contract docs for more details.
 contract ERC20DripsHub is ManagedDripsHub {
-    /// @notice The ERC-1967 storage slot for the contract.
-    /// It holds a single address of the ERC-20 reserve.
-    bytes32 private constant SLOT_RESERVE =
-        bytes32(uint256(keccak256("eip1967.erc20DripsHub.reserve")) - 1);
     /// @notice The address of the ERC-20 contract which tokens the drips hub works with
     IERC20 public immutable erc20;
-
-    /// @notice Emitted when the reserve address is set
-    event ReserveSet(IERC20Reserve oldReserve, IERC20Reserve newReserve);
+    /// @notice The address of the ERC-20 reserve which the drips hub works with
+    IERC20Reserve public immutable reserve;
 
     /// @param cycleSecs The length of cycleSecs to be used in the contract instance.
     /// Low value makes funds more available by shortening the average time of funds being frozen
     /// between being taken from the users' drips balances and being collectable by their receivers.
     /// High value makes collecting cheaper by making it process less cycles for a given time range.
     /// @param _erc20 The address of an ERC-20 contract which tokens the drips hub will work with.
-    constructor(uint64 cycleSecs, IERC20 _erc20) ManagedDripsHub(cycleSecs) {
+    /// @param _reserve The address of the ERC-20 reserve which the drips hub will work with
+    constructor(
+        uint64 cycleSecs,
+        IERC20 _erc20,
+        IERC20Reserve _reserve
+    ) ManagedDripsHub(cycleSecs) {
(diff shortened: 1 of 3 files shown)


deckardkrug:
https://github.com/radicle-dev/radicle-drips-hub/pull/92/files#diff-8e809381fe4e2176c4ff79e575602d34a22fc9a2511f3053a00b883978d0e285R19

Why do we call `reserve.setUser(address(dripsHub));` instead of passing the user in the constructor?

I understand that we need the address from the drips-hub but we could calculate the address as done [here](https://github.com/radicle-dev/radicle-contracts/blob/master/contracts/deploy/phase0.sol#L44-L46) and then pass it like this `ERC20Reserve reserve = new ERC20Reserve(erc20, address(this), address(dripsHub));`.

Then we could also remove `setUser()` and make the user immutable.

CodeSandwich:
@deckardkrug This is because the reserve needs to have `setUser` anyway. The idea is that in case of a catastrophic failure we can first cut off drips hub and then deploy a refund contract and set it as the user instead of drips hub without giving away the admin role. With `setUser` in place we don't need to do any fancy logic to predict the addresses, we can use the clearest and hardest to mess up approach.

deckardkrug:
@CodeSandwich okay - thank you for clarifying!

xmxanuel:
@deckardkrug I think it is a potential interesting idea but still a bit risky, however it might be cleaner with factory contract and the `create2` opcode.  

As Igor mentioned we want to have the `setUser` method.

xmxanuel:
@CodeSandwich The only potential use case I can see if the reserve has a bug and we are migrating the reserve itself.
 
 In that worst case scenario, however we could also upgrade the contract and define it with a new constant.

src/ERC20DripsHub.sol
```diff
@@ -127,37 +128,16 @@ contract ERC20DripsHub is ManagedDripsHub {
...
         if (amt > 0) {
             uint256 withdraw = uint128(amt);
-            erc20Reserve.withdraw(withdraw);
+            reserve.withdraw(withdraw);
             erc20.transfer(user, withdraw);
         } else if (amt < 0) {
             uint256 deposit = uint128(-amt);
             erc20.transferFrom(user, address(this), deposit);
-            erc20Reserve.deposit(deposit);
+            erc20.approve(address(reserve), deposit);
```
xmxanuel, line 139: It will be a bit more expensive to approve every time instead of the `approve uint.max`. 

However, I see the problem we would need some `register` function for first time usage of erc20. 

One idea could be have a blacklist of erc20 controlled by governance. The register function could check if an erc20 is the blacklist otherwise the approval happens here.

xmxanuel (approved):
lgtm

src/ERC20DripsHub.sol
CodeSandwich: Yeah, that's a solution to the unseen token approvals. We could check approved amount and set it to `uint.max` if it's different. That would probably remove a need for blacklists, IMO we don't need to protect from stealcoins with malicious implementations, if the users want to use it and they get rolled or the whole drips hub is stuck for that token, it's their responsibility. My plan here is to keep it as simple as possible because I want the funds to be transferred directly from the drivers to the reserve, so this code can be removed anyway.

It's not that much more expensive, in both cases the ERC20's `approved[dripsHub][reserve]` doesn't change. With `uint.max` that's true for all sane ERC-20 implementations where this value isn't updated (the standard doesn't enforce it). With `0 -> <deposit> -> 0` all storage costs are refunded because in the end nothing is changed. Sure, refunds aren't as convenient and reliable as not being charged in the first place.

xmxanuel:
@CodeSandwich #101 I am wondering if this might be a reason to change the reserve later

CodeSandwich:
IMO no, changing the reserve address would also mean moving funds spread across an unlimited number of ERC-20s, which is extremely difficult. However we implement plugins, the reserve will probably have a constant address.

PR #93, opened 2022-02-14
Make reserve multi-token

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/95. Blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/92

CC: [deckardkrug](https://github.com/deckardkrug)


src/ERC20Reserve.sol
```diff
@@ -5,54 +5,49 @@ import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
 
 interface IERC20Reserve {
-    function erc20() external view returns (IERC20);
+    function withdrawable(IERC20 token) external view returns (uint256 amt);
```
xmxanuel, line 8: I still would prefer to use `address` here. It just results in unnecessary casting which is anyway not enforced.
We had the discussion already. :)

src/ERC20Reserve.sol
```diff
@@ -5,54 +5,49 @@ import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
...
-    function withdraw(uint256 amt) public override onlyUser {
+    function withdraw(IERC20 token, uint256 amt) public override onlyUser {
+        uint256 balance = withdrawable[token];
         require(balance >= amt, "Reserve: withdrawal over balance");
-        balance -= amt;
-        emit Withdrawn(_msgSender(), amt);
-        require(erc20.transfer(_msgSender(), amt), "Reserve: transfer failed");
+        withdrawable[token] = balance - amt;
+        require(token.transfer(msg.sender, amt), "Reserve: transfer failed");
+        emit Withdrawn(msg.sender, amt);
```
xmxanuel, line 39: I think we should add the erc20 `address`  to the deposit, withdraw events

src/ERC20Reserve.sol
xmxanuel: We can keep it.

src/ERC20Reserve.sol
```diff
@@ -5,54 +5,49 @@ import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
...
+        withdrawable[token] += amt;
+        require(token.transferFrom(msg.sender, address(this), amt), "Reserve: transfer failed");
+        emit Deposited(msg.sender, amt);
     }
 
-    function forceWithdraw(uint256 amt) public onlyOwner {
-        emit ForceWithdrawn(_msgSender(), amt);
-        require(erc20.transfer(_msgSender(), amt), "Reserve: transfer failed");
+    function forceWithdraw(IERC20 token, uint256 amt) public onlyOwner {
+        require(token.transfer(msg.sender, amt), "Reserve: transfer failed");
```
xmxanuel, line 49: Mhm, I think we could reduce the withdrawable here but allow to withdraw more than the balance.
```solidity
if(balance <= amt) {
 withdrawable[token] = 0;
} else {
 withdrawable[token] = balance - amt;
}
```

src/ERC20Reserve.sol
```diff
@@ -5,54 +5,49 @@ import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
...
-    function withdraw(uint256 amt) external;
+    function withdraw(IERC20 token, uint256 amt) external;
 
-    function deposit(uint256 amt) external;
+    function deposit(IERC20 token, uint256 amt) external;
 }
 
 contract ERC20Reserve is IERC20Reserve, Ownable {
-    IERC20 public immutable override erc20;
     address public user;
```
xmxanuel, line 16: I think i still would prefer a mapping for the `user`. It would just give us more potential freedom, in case we end up with a legacy version which still wants to have access to the reserve and a newer one etc.

xmxanuel (commented):
looks good. :+1:  
mostly some thoughts about the general reserve design.

The only thing which I think we should add are the addresses of the erc20s to the events

src/ERC20Reserve.sol
CodeSandwich: One of the use cases for forced withdrawal is to shave off tokens which are in the reserve but will never be withdrawn, e.g. the DSR-generating income. If we reduced `withdrawable`, the drips hub could get stuck.

Forced withdrawal is a counterpart of ERC-20 transfers, they too allow changing the actual balance of the reserve without touching `withdrawable`.

src/ERC20Reserve.sol
CodeSandwich: I agree, fixed

src/ERC20Reserve.sol
CodeSandwich: Fixed

src/ERC20Reserve.sol
xmxanuel: Ah, fair point to keep it like that. Maybe we could call the function `harvest` which only allows to get the difference between `erc20.balanceOf` and the stored `balance`. 

If you want all of it, you could still make yourself a regular user.

CodeSandwich: Yeah, adding `harvest` could make sense :thinking:

Pros:
- More difficult for `forceWithdraw` to make the reserve go below `withdrawable` by mistake
Cons:
- More complex API with more logic and weirder, harder to use functions

`harvest` can be emulated using `forceWithdraw` as:
```solidity
function harvest(uint256 amt) {
    require(amt < token.balanceOf(address(reserve)) - reserve.withdrawable(token));
    reserve.forceWithdraw(amt);
}

PR #94, opened 2022-02-14
feat: storage functions in assembly

xmxanuel:
I was curious how much gas we can safe, if we perform the storage operations directly in assembly.

The changes are not that big but actually a tiny bit cheaper.
I think you could also optimize this further with directly accessing the  `userOrAccount` in memory.

Might be interesting for the delta storage operations but I think there would also have to do some bit manipulation for the `uint128` shared slots.

Docs: https://docs.soliditylang.org/en/v0.8.11/internals/layout_in_storage.html?highlight=storage


CodeSandwich:
Interesting! IIUC this is basically exactly what Solidity is doing but in yul? How much gas have you managed to shave off?

PR #102, opened 2022-02-16
Forge migration

deckardkrug:


foundry.toml
```diff
@@ -0,0 +1,15 @@
+[default]
+src = 'src'
+out = 'out'
+libs = ['lib']
+solc-version = "0.8.7"
+optimizer = true
+optimizer_runs = 1000000
+verbosity = 1
+fuzz_runs = 5
```
deckardkrug, line 9: I suggest we increase `fuzz_runs`, as forge does fuzzing by [default](https://github.com/gakonst/foundry/tree/master/forge#fuzzing-go-beyond-unit-testing).

deckardkrug:
Ready to review.

foundry.toml
CodeSandwich: I think that 5 is a good default value. It's used when running tests locally, so increasing it too much would slow down the developer flow. The CI is a different story, it has time to crunch a lot of runs.

foundry.toml
```diff
@@ -0,0 +1,15 @@
...
+src = 'src'
+out = 'out'
+libs = ['lib']
+solc-version = "0.8.7"
+optimizer = true
+optimizer_runs = 1000000
+verbosity = 1
+fuzz_runs = 5 
+remappings = [
+    'ds-test/=lib/ds-test/src/',
```
CodeSandwich, line 11: `forge-std` has a dependency on `ds-test`: https://github.com/brockelmore/forge-std/tree/master/lib, can we reuse it with something like `'ds-test/=lib/forge-std/lib/ds-test/src'`? Having two version in a project is a ticking bomb. I also wouldn't be surprised if forge created its own version, the original has some aspect to improve and it may need to start using the new cheatcodes.

foundry.toml
```diff
@@ -0,0 +1,15 @@
+[default]
+src = 'src'
+out = 'out'
+libs = ['lib']
+solc-version = "0.8.7"
+optimizer = true
```
CodeSandwich, line 6:
```suggestion
solc-version = "0.8.7"
```
Most of these entries are just copied forge defaults and they don't improve readability by being explicit, IMO we could drop them for cleanness.

Makefile
```diff
@@ -1,12 +1,10 @@
-install     : install_solc dapp_update yarn_install
-install_solc:; nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-static-versions.solc_0_8_7
-dapp_update :; dapp update
+install     : forge_update yarn_install
```
CodeSandwich, line 1: This means that the simple `make install` run by a developer **or the CI** will pull the newest versions of all the dependencies, which sounds like an endless source of fun if an update breaks anything :D

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "lib/openzeppelin-contracts"]
 	path = lib/openzeppelin-contracts
 	url = https://github.com/OpenZeppelin/openzeppelin-contracts
```
CodeSandwich, line 6: Is openzeppelin upgrade necessary? Why to `master` and not to one of official releases? I'm fine with the upgrade, but maybe it'd make sense to make it a separate PR not to mix potential issues and to make it perfectly clear what's happening.

.github/workflows/build.yml
```diff
@@ -8,19 +8,19 @@ jobs:
...
+        with:
+          submodules: recursive
       - uses: actions/setup-node@v2
-      - uses: cachix/install-nix-action@v13
-      - uses: cachix/cachix-action@v10
+      - uses: onbjerg/foundry-toolchain@v1
         with:
-          name: radicle-drips-hub
-          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          version: nightly
```
CodeSandwich, line 16: Maybe we should stick to a specific nightly version, the one with the `-<hash>` suffix. Otherwise the CI will run a different version every few days.

CodeSandwich (changes requested):
Looks really nice, good work!

foundry.toml
deckardkrug: Good point.

foundry.toml
deckardkrug: Okay, I will remove them.

Makefile
deckardkrug: You are right. That is nothing we want, however isn't `dapp update` doing the same?

.github/workflows/build.yml
deckardkrug: Agree.

Makefile
deckardkrug: Okay. As it makes sense to do updates manually instead of running tests against an updated version the code was not developed with, I will remove this command from the ci.

foundry.toml
deckardkrug: I removed all default entries and only added the ones that need to be overridden by the entries that are specified in the current `.dapprc`.

.github/workflows/build.yml
```diff
@@ -8,19 +8,19 @@ jobs:
...
       - name: lint
-        run: nix-shell --run "make lint"
+        run: make lint 
+
       - name: test
         if: ${{ github.event_name == 'pull_request' }}
-        run: nix-shell --run "make test"
-      - name: test deep
-        if: ${{ github.event_name == 'push' }}
-        run: nix-shell --run "make test_deep"
```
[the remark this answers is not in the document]
deckardkrug: As forge test does not accept a `--fuzz-runs` flag, I have established a `ci` profile in `foundry.toml` which is used in the ci.

Makefile
xmxanuel: `dapp upgrade` would update the dependencies to the latest version in dapp.tools

xmxanuel: Seems like the changed it for forge. However I agree `install` and `update` is more clear than `update` and `upgrade`.

foundry.toml
```diff
@@ -0,0 +1,21 @@
...
+    'ds-test/=lib/forge-std/lib/ds-test/src/',
+    'openzeppelin-contracts/=lib/openzeppelin-contracts/contracts/'
+]
+
+[ci]
+solc_version = "0.8.7"
+optimizer_runs = 1000000
+verbosity = 1
+fuzz_runs = 50000
+remappings = [
```
xmxanuel, line 17: In this template the remappings are in an individual file.

https://github.com/FrankieIsLost/forge-template

I don't have a strong opinion. Maybe we can just follow the standard. (If there is already a common approach how to do it. :) )

xmxanuel (approved):
nice work @deckardkrug

foundry.toml
CodeSandwich: That repo doesn't even have `foundry.toml` :thinking: I wouldn't be surprised if it used an old version of foundry, it evolves really fast!

xmxanuel: Ah, I think you might be right. 👍

deckardkrug: Right, the `foundry.toml` file was established recently.

.github/workflows/build.yml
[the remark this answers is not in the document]
CodeSandwich: The problem isn't resolved, as of now `master` still doesn't run more CI runs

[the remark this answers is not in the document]
deckardkrug: It seems like this is a bug then or do you think the ci resets the environment variables for each run?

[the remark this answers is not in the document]
deckardkrug: I just added the env variable manually, let's try if that solves the issue.

.github/workflows/build.yml
```diff
@@ -3,24 +3,28 @@ on:
   pull_request:
   push:
     branches: [ master ]
+
+env:
+  FOUNDRY_PROFILE: ci
```
CodeSandwich, line 8: According to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env it sets the `FOUNDRY_PROFILE` for all jobs making the PR runs as long as the `master` runs. This is unfortunate because work on the PRs is often slowed down by the CI. This is why there were 2 different configurations: a quick one for the PRs and a long one for `master` after the PR is merged and nobody waits for the CI to finish anymore. If we set that variable in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsenv, we can restore that setup.

foundry.toml
```diff
@@ -0,0 +1,21 @@
+[default]
+solc_version = "0.8.7"
```
CodeSandwich, line 2: Something's off, the CI says that it compiles contracts using Solidity 0.8.12, but THB I don't understand why?

.github/workflows/build.yml
CodeSandwich: They must've added it like today but fuzz runs can now be passed using the env variable: https://onbjerg.github.io/foundry-book/reference/config.html#fuzz_runs If it overrides `foundry.toml` (it should), then we don't need a separate profile.

foundry.toml
deckardkrug: I just noticed that this also happens on my local machine. This is probably foundry bug - I will investigate that more.

.github/workflows/build.yml
```diff
@@ -4,13 +4,14 @@ on:
...
-env:
-  FOUNDRY_PROFILE: ci 
-
 jobs:
   tests:
     runs-on: ubuntu-latest
     steps:
+      - name: Environment setup
+        env:
+          FOUNDRY_FUZZ_RUNS: 50000
```
CodeSandwich, line 13: I think that it applies only to the step where it's defined, so it needs to be added to `dapp test`

foundry.toml
deckardkrug: Seems like setting `FOUNDRY_SOLC_VERSION` does not work either.

.github/workflows/build.yml
deckardkrug: I might just have overseen that as I was only searching for a flag that I can pass to the command.

.github/workflows/build.yml
```diff
@@ -3,24 +3,27 @@ on:
...
+
       - name: test
         if: ${{ github.event_name == 'pull_request' }}
-        run: nix-shell --run "make test"
-      - name: test deep
-        if: ${{ github.event_name == 'push' }}
-        run: nix-shell --run "make test_deep"
+        env:
+          FOUNDRY_FUZZ_RUNS: 50000
+        run: make test
```
CodeSandwich, line 29: Almost there, now the long fuzz runs are run for the PR CI runs and after merging to `master` nothing is run at all.

xmxanuel:
@CodeSandwich I think it would be still nice to migrate at a certain point.  However, I think it might be easier after we got all the changes from the next Drips version merged.

PR #104, opened 2022-02-18
Move ERC-1967 drips hub storage logic to ManagedDripsHub

CodeSandwich:
Improve separation of concerns by pushing the upgrade-specific logic from DripsHub to ManagedDripsHub.

### DIFF
--- a/src/ManagedDripsHub.sol
+++ b/src/ManagedDripsHub.sol
@@ -18,9 +18,9 @@ import {DripsHub, SplitsReceiver} from "./DripsHub.sol";
 /// When a proxy uses such contract via delegation, it's initially unpaused.
 abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable {
     /// @notice The ERC-1967 storage slot for the contract.
-    /// It holds a single boolean indicating if the contract is paused.
-    bytes32 private constant SLOT_PAUSED =
-        bytes32(uint256(keccak256("eip1967.managedDripsHub.paused")) - 1);
+    /// It holds a single `ManagedDripsHubStorage` structure.
+    bytes32 private constant STORAGE_SLOT =
+        bytes32(uint256(keccak256("eip1967.managedDripsHub.storage")) - 1);
 
     /// @notice Emitted when the pause is triggered.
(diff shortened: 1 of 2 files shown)


src/ManagedDripsHub.sol
```diff
@@ -128,9 +133,30 @@ abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable {
...
+        pure
+        returns (ManagedDripsHubStorage storage storageRef)
+    {
+        bytes32 slot = STORAGE_SLOT;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            // Based on OpenZeppelin's StorageSlot
+            storageRef.slot := slot
+        }
     }
```
deckardkrug, line 160: @CodeSandwich What do you think about making `DripsHub` itself `UUPSUpgradeable` so we can remove `ManagedDripsHub` completely?

As I see it, the only reason that would speak for not doing that would be that we need to be able to deploy a `DripsHub` instance without upgradeability.

deckardkrug: Agree.

> maybe even abstract DripsHub further and extract from it a few simpler libraries.

So, we could do a `SplitsLibrary` and `DripsLibrary`.

src/ManagedDripsHub.sol
```diff
@@ -128,9 +133,30 @@ abstract contract ManagedDripsHub is DripsHub, UUPSUpgradeable {
...
-        return StorageSlot.getBooleanSlot(SLOT_PAUSED);
+    /// @notice Returns the DripsHub storage.
+    /// @return storageRef The storage.
+    function _dripsHubStorage()
+        internal
+        view
+        override
+        returns (DripsHubStorage storage storageRef)
+    {
+        return _managedDripsHubStorage().dripsHubStorage;
```
xmxanuel, line 144: I think it is a good idea to just use one storage slot for managedDripsHub and the actual dripsHub.

Since the `dripsHubStorage` is at the first position in the struct.  We could check if it gets cheaper to just return the slot in assembly here. Instead of calling the method.

xmxanuel (approved):
Looks good.

PR #105, opened 2022-02-22
Restructure DripsHub storage

CodeSandwich:
Prepares the storage for the upcoming changes like multi-token, drivers and refactoring into libs.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -151,27 +151,6 @@ abstract contract DripsHub {
         uint128 amt
     );
 
-    struct ReceiverState {
-        // The amount collectable independently from cycles
-        uint128 collectable;
-        // The next cycle to be collected
-        uint64 nextCollectedCycle;
-        // --- SLOT BOUNDARY
-        // The changes of collected amounts on specific cycle.
-        // The keys are cycles, each cycle `C` becomes collectable on timestamp `C * cycleSecs`.
-        // Values for cycles before `nextCollectedCycle` are guaranteed to be zeroed.
-        // This means that the value of `amtDeltas[nextCollectedCycle].thisCycle` is always
-        // relative to 0 or in other words it's an absolute value independent from other cycles.
-        mapping(uint64 => AmtDelta) amtDeltas;
-    }
-
-    struct AmtDelta {
-        // Amount delta applied on this cycle
-        int128 thisCycle;
-        // Amount delta applied on the next cycle
-        int128 nextCycle;
-    }
-
     struct UserOrAccount {
         bool isAccount;
         address user;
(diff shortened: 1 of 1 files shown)


src/DripsHub.sol
```diff
@@ -151,46 +152,48 @@ abstract contract DripsHub {
...
+    struct UserOrAccount {
+        bool isAccount;
+        address user;
+        uint256 account;
+    }
+
+    struct DripsHubStorage {
+        /// @notice User drips states.
+        /// The keys are the user hash and the asset ID.
+        mapping(bytes32 => mapping(uint256 => DripsState)) dripsStates;
```
xmxanuel, line 164: Why is the asset ID a uint and not an address?

I mean technically it is anyway the hash of it, so it doesn't really matter.

src/DripsHub.sol
```diff
@@ -219,21 +222,24 @@ abstract contract DripsHub {
         view
         returns (uint128 collected, uint128 split)
     {
-        ReceiverState storage receiver = _dripsHubStorage().receiverStates[user];
+        bytes32 userHash = _hashUser(user);
```
xmxanuel, line 225: I think `userID` would be a better name. Also maybe `_calcUserID` instead of `_hashUser`.

Then it is clearer, that is an `ID`. Otherwise the reader might confuse it with all the other hashes we are dealing with.

src/DripsHub.sol
```diff
@@ -219,21 +222,24 @@ abstract contract DripsHub {
...
+        DripsHubStorage storage dripsHubStorage = _dripsHubStorage();
+        DripsState storage dripsState = dripsHubStorage.dripsStates[userHash][DUMMY_ASSET];
+
         _assertCurrSplits(user, currReceivers);
 
         // Collectable independently from cycles
-        collected = receiver.collectable;
+        collected = dripsHubStorage.splitsStates[userHash].collectables[DUMMY_ASSET];
```
xmxanuel, line 232: I think it would be cleaner
```solidity
uint collected = 0;
// collect logic
collected += dripsHubStorage.splitsStates[userHash].collectables[DUMMY_ASSET]
// split logic
```

src/DripsHub.sol
```diff
@@ -828,6 +829,22 @@ abstract contract DripsHub {
...
+        }
+    }
+
+    function _hashUser(address user) internal pure returns (bytes32) {
+        return keccak256(abi.encode(user));
+    }
+
+    function _hashUser(address user, uint256 account) internal pure returns (bytes32) {
+        return keccak256(abi.encode(user, account));
+    }
```
xmxanuel, line 846: I would make both functions public. It just easier of another contract wants to build on top and can use these functions.

src/DripsHub.sol
```diff
@@ -828,6 +829,22 @@ abstract contract DripsHub {
         return UserOrAccount({isAccount: true, user: user, account: account});
     }
 
+    function _hashUser(UserOrAccount memory userOrAccount) internal pure returns (bytes32) {
```
xmxanuel, line 832: This will soon be legacy and can be removed, correct?

src/DripsHub.sol
CodeSandwich: On the abstract DripsHub level the token address doesn't really matter, it's just a mapping key and an event field, so it could be anything capable of telling apart 2^20 values. If we want to go in the library direction, we'll need to make it as universal as possible and not assume that the asset is an ERC-20 token identified by its address, this is the role of the wrapping contract implementing a specific API. I think that in this context uint256 will be the right choice.

So should we stick with `uint256`, tighten it down to `address` or make even bigger assumptions and make it `IERC20`? I think that `uint256` is both the cleanest and the most aligned with our future plans preventing disruptions. WDYT?

src/DripsHub.sol
CodeSandwich: Correct!

src/DripsHub.sol
CodeSandwich: `collected` is a returned variable, so it's already initialized to 0, writing `collected = 0;` to me just doesn't feel right. Would only turning `=` into `+=` be clear enough?

```solidity
collected += dripsHubStorage.splitsStates[userHash].collectables[DUMMY_ASSET];
```

src/DripsHub.sol
CodeSandwich: Agreed, fixed

src/DripsHub.sol
CodeSandwich: That's a good point, fixed

src/DripsHub.sol
deckardkrug: I like the idea of abstracting it more by using a `uint256`. This would allow other use cases e.g. for games that use a unique point system or for applications that are similar to how current real world reward systems work that don't want to have a tradable token.

I can also imagine that this would be useful for a later stage when we want to add anonymity to the protocol.

Issue #96, opened 2022-02-15
Add support for multiple tokens to hub


xmxanuel:
- stop setSplits from collecting

PR #107, opened 2022-02-23
Almost multi-token DripsHub

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-drips-hub/issues/96, blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/105.

Stops `setSplits` from collecting. Adds asset ID to events, as of now it's just `DUMMY_ASSET` and it's unindexed until we remove support for raw address user identity and free up indexed event slots.

The next step is to add the actual asset ID to the DripsHub API.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -69,6 +69,7 @@ abstract contract DripsHub {
     /// and`endTime` (exclusively) or until the timestamp of the next drips update (exclusively).
     /// @param user The dripping user
     /// @param receiver The receiver of the updated drips
+    /// @param assetId The used asset ID
     /// @param amtPerSec The new amount per second dripped from the user
     /// to the receiver or 0 if the drips are stopped
     /// @param endTime The timestamp when dripping will stop,
@@ -76,6 +77,7 @@ abstract contract DripsHub {
     event Dripping(
         address indexed user,
         address indexed receiver,
+        uint256 assetId,
         uint128 amtPerSec,
         uint64 endTime
     );
@@ -86,6 +88,7 @@ abstract contract DripsHub {
     /// @param user The user
     /// @param account The dripping account
     /// @param receiver The receiver of the updated drips
+    /// @param assetId The used asset ID
     /// @param amtPerSec The new amount per second dripped from the user's account
     /// to the receiver or 0 if the drips are stopped
     /// @param endTime The timestamp when dripping will stop,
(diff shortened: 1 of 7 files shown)


src/DripsHub.sol
```diff
@@ -638,9 +655,16 @@ abstract contract DripsHub {
...
+                userOrAccount.user,
+                userOrAccount.account,
+                receiver,
+                DUMMY_ASSET,
+                amtPerSec,
+                endTime
+            );
         } else {
-            emit Dripping(userOrAccount.user, receiver, amtPerSec, endTime);
+            emit Dripping(userOrAccount.user, receiver, DUMMY_ASSET, amtPerSec, endTime);
```
xmxanuel, line 667: Do you think we can remove the `userOrAccount` struct? Maybe only emitting events on a higher layer?

src/DripsHub.sol
```diff
@@ -714,27 +738,16 @@ abstract contract DripsHub {
...
-    ) internal returns (uint128 collected, uint128 split) {
-        (collected, split) = _collectInternal(user, currReceivers);
-        _assertSplitsValid(newReceivers);
-        _dripsHubStorage().splitsStates[calcUserId(user)].splitsHash = hashSplits(newReceivers);
-        emit SplitsUpdated(user, newReceivers);
-        _transfer(user, int128(collected));
+    function _setSplits(address user, SplitsReceiver[] memory receivers) internal {
+        _assertSplitsValid(receivers);
+        _dripsHubStorage().splitsStates[calcUserId(user)].splitsHash = hashSplits(receivers);
+        emit SplitsUpdated(user, receivers);
```
xmxanuel, line 750: nice, this is now a cheaper operation

src/DripsHub.sol
```diff
@@ -339,11 +356,11 @@ abstract contract DripsHub {
                 split += splitsAmt;
                 address splitsReceiver = currReceivers[i].receiver;
                 splitsStates[calcUserId(splitsReceiver)].collectables[DUMMY_ASSET] += splitsAmt;
-                emit Split(user, splitsReceiver, splitsAmt);
+                emit Split(user, splitsReceiver, DUMMY_ASSET, splitsAmt);
```
xmxanuel, line 359: We also need to move `splitting part` to an own function and call it here. 

I guess this will happen in the upcoming PRs.

xmxanuel (approved):
looks good 🚀

src/DripsHub.sol
CodeSandwich: I was planning to get rid of it when we no longer allow non-account identity, until then it's still somewhat needed and probably not worth bigger refactorings.

src/DripsHub.sol
CodeSandwich: All in its own time, but you're right, we'll do that too :+1:

PR #108, opened 2022-02-25
Expose asset ID in the APIs

CodeSandwich:
The final step to close https://github.com/radicle-dev/radicle-drips-hub/issues/96.

### DIFF
--- a/src/DaiDripsHub.sol
+++ b/src/DaiDripsHub.sol
@@ -24,7 +24,7 @@ contract DaiDripsHub is ERC20DripsHub {
         uint64 cycleSecs,
         IDai _dai,
         IERC20Reserve reserve
-    ) ERC20DripsHub(cycleSecs, _dai, reserve) {
+    ) ERC20DripsHub(cycleSecs, reserve) {
         dai = _dai;
     }
 
@@ -44,7 +44,15 @@ contract DaiDripsHub is ERC20DripsHub {
         PermitArgs calldata permitArgs
     ) public whenNotPaused returns (uint128 newBalance, int128 realBalanceDelta) {
         _permit(permitArgs);
-        return setDrips(lastUpdate, lastBalance, currReceivers, balanceDelta, newReceivers);
+        return
+            setDrips(
+                uint160(address(dai)),
+                lastUpdate,
+                lastBalance,
+                currReceivers,
+                balanceDelta,
+                newReceivers
+            );
     }
 
     /// @notice Sets the drips configuration of an account of the `msg.sender`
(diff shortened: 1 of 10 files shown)


src/DaiDripsHub.sol
```diff
@@ -24,7 +24,7 @@ contract DaiDripsHub is ERC20DripsHub {
         uint64 cycleSecs,
         IDai _dai,
         IERC20Reserve reserve
-    ) ERC20DripsHub(cycleSecs, _dai, reserve) {
+    ) ERC20DripsHub(cycleSecs, reserve) {
```
xmxanuel, line 27: Maybe we should support the general `permit` standard as well.

src/DaiDripsHub.sol
```diff
@@ -81,7 +89,7 @@ contract DaiDripsHub is ERC20DripsHub {
         PermitArgs calldata permitArgs
     ) public whenNotPaused {
         _permit(permitArgs);
-        give(receiver, amt);
+        give(receiver, uint160(address(dai)), amt);
```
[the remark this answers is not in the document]
xmxanuel: It is also a bit confusing it actually should be `uint256(uint160(address(dai)))`. 

Is `uint256((address(dai))` resulting in a compiler warning?

src/DripsHub.sol
```diff
@@ -355,26 +362,26 @@ abstract contract DripsHub {
...
     /// @param user The user
+    /// @param assetId The used asset ID
     /// @param count The number of flushed cycles.
     /// @return collectedAmt The collected amount
-    function _flushCyclesInternal(address user, uint64 count)
-        internal
-        returns (uint128 collectedAmt)
-    {
+    function _flushCyclesInternal(
+        address user,
```
xmxanuel, line 379: I think it might be worth consider changing the pattern a bit. We could already pass here the `bytes32 userId` instead of the address.

We could calculate the userId on that layer where we deal with `msg.sender`, etc.

I think this would be a bit cleaner. Then the lower layer isn't even aware that a userId is calculated from an address.

xmxanuel: Maybe you are planning to do this anyway as part of the `userOrAccount` refactoring.

xmxanuel (approved):
good work. 💯 

I added some potential ideas.

src/DripsHub.sol
CodeSandwich: I agree, that's the goal, but let's do one thing at a time, this PR is already almost 700 line changes.

src/DaiDripsHub.sol
CodeSandwich: I agree, that will be the responsibility of the drivers

src/DaiDripsHub.sol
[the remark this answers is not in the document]
CodeSandwich: Yes, casting from `address` to `uint256` is forbidden in modern Solidity. The logic is that you may expect the casting to choose the `address -> bytes20 -> bytes32 -> uint256` route which would result in different padding. Casting uint160 -> uint256 is implicit.

The goal is to not have to cast `uint256 -> IERC20`, the current implementation isn't finished yet, it's a minimal set of changes.

Issue #97, opened 2022-02-15
low level functions for collect and split


xmxanuel:
- collect is callable only by the collected address itself.

PR #109, opened 2022-03-07
Prepare for splitting without collecting API

CodeSandwich:
First part of https://github.com/radicle-dev/radicle-drips-hub/issues/97. Adjusts the existing API for splitting without collecting but doesn't add any new functions yet.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -239,7 +239,7 @@ abstract contract DripsHub {
     /// @param currReceivers The list of the user's current splits receivers.
     /// @return collected The collected amount
     /// @return split The amount split to the user's splits receivers
-    function collectable(
+    function collectableAll(
         address user,
         uint256 assetId,
         SplitsReceiver[] memory currReceivers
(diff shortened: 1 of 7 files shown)


src/DripsHub.sol
```diff
@@ -204,11 +204,17 @@ abstract contract DripsHub {
...
-        mapping(uint256 => uint128) collectables;
+        /// @notice The user's splits balance. The key is the asset ID.
+        mapping(uint256 => SplitsBalance) balances;
+    }
+
+    struct SplitsBalance {
+        /// @notice The not yet split balance, must be split before collecting by the user.
+        uint128 unsplit;
+        /// @notice The already split balance, ready to be collected by the user.
+        uint128 split;
```
xmxanuel, line 217: So, basically the `split` is kind of an `accountBalance`.

Are you envisioning the `receiveDrips` to be callable  as standalone?
If thats the case, we would have to increase this balance here in the splitStruct.

I have the feeling the variable should live somewhere else, however for gas reasons it makes sense to put it next to `unsplit`.

Maybe the `SplitBalance` struct is more like an `AccountBalance` with a `balance` and `unsplit`

xmxanuel (approved):
looks good

PR #111, opened 2022-03-14
Make collecting for a user callable only by the user

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/97, blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/110.


xmxanuel (approved):
looks good

PR #112, opened 2022-03-15
Replace user accounts with user IDs

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-drips-hub/issues/98, blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/111.

Adds the notion of a `userId` composed of the `account` and `subAccount`. Replaces all usage of accounts (address + uint256) with the new userId (uint256). It's been used for the user identification only in 2 functions: in `setDrips` (only the sender) and `give` (only the giver).

Temporarily adds the special account 0, which allows controlling sub-account equal to `msg.sender` without account creation. This is needed so we can keep the old API implicitly using `msg.sender` to identify the users, e.g. in `function give(address receiver, uint128 amt)`. This special case can be dropped when we switch the whole API to `userId` and outsource the EOA API to a separate contract.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -198,6 +198,10 @@ abstract contract DripsHub {
         /// @notice User splits states.
         /// The key is the user ID.
         mapping(uint256 => SplitsState) splitsStates;
+        /// @notice The last created account ID. The next one will have ID lastAccountId + 1.
+        uint32 lastAccountId;
+        /// @notice Account owners. The key is the account ID, the value is the owner address.
+        mapping(uint32 => address) accountsOwners;
     }
 
     struct DripsState {
@@ -240,6 +244,30 @@ abstract contract DripsHub {
     /// High value makes collecting cheaper by making it process less cycles for a given time range.
     constructor(uint64 _cycleSecs) {
         cycleSecs = _cycleSecs;
+    }
+
+    /// @notice Creates an account.
+    /// Assigns it an ID and lets its owner perform actions on behalf of all its sub-accounts.
+    /// Multiple accounts can be registered for a single address, it will own all of them.
+    /// @return accountId The new account ID.
+    function createAccount(address owner) public virtual returns (uint32 accountId) {
+        DripsHubStorage storage dripsHubStorage = _dripsHubStorage();
+        accountId = dripsHubStorage.lastAccountId + 1;
+        dripsHubStorage.lastAccountId = accountId;
+        dripsHubStorage.accountsOwners[accountId] = owner;
+    }
+
+    /// @notice Returns account owner.
+    /// @param accountId The account to look up.
+    /// @return owner The owner of the account. If the account doesn't exist, returns address 0.
+    function accountOwner(uint32 accountId) public view returns (address owner) {
+        return _dripsHubStorage().accountsOwners[accountId];
+    }
+
+    /// @notice Returns the ID which will be assigned for the next created account.
+    /// @return accountId The account ID.
+    function nextAccountId() public view returns (uint32 accountId) {
+        return _dripsHubStorage().lastAccountId + 1;
     }
 
     /// @notice Returns the contract storage.
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -62,11 +62,13 @@ abstract contract DripsHub {
     uint32 public constant MAX_SPLITS_RECEIVERS = 200;
     /// @notice The total splits weight of a user
     uint32 public constant TOTAL_SPLITS_WEIGHT = 1_000_000;
+    /// @notice Number of bits in the sub-account part of userId
+    uint256 public constant BITS_SUB_ACCOUNT = 224;
 
     /// @notice Emitted when drips from a user to a receiver are updated.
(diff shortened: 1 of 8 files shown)


src/DripsHub.sol
```diff
@@ -242,6 +192,47 @@ abstract contract DripsHub {
...
+    /// @notice Creates an account.
+    /// Assigns it an ID and lets its owner perform actions on behalf of all its sub-accounts.
+    /// Multiple accounts can be registered for a single address, it will own all of them.
+    /// @return accountId The new account ID.
+    function createAccount(address owner) public virtual returns (uint32 accountId) {
+        DripsHubStorage storage dripsHubStorage = _dripsHubStorage();
+        accountId = dripsHubStorage.lastAccountId + 1;
+        dripsHubStorage.lastAccountId = accountId;
+        dripsHubStorage.accountsOwners[accountId] = owner;
+    }
```
xmxanuel, line 221: nice, I like it. Actually very clean.

src/DripsHub.sol
```diff
@@ -242,6 +192,47 @@ abstract contract DripsHub {
         cycleSecs = _cycleSecs;
     }
 
+    modifier onlyAccountOwner(uint256 userId) {
+        uint32 accountId = uint32(userId >> 224);
```
xmxanuel, line 196: Lets use a constant for it.
```suggestion
        uint32 accountId = uint32(userId >> BITS_SUB_ACCOUNT);
```

src/DripsHub.sol
```diff
@@ -471,22 +462,18 @@ abstract contract DripsHub {
...
     /// @param assetId The used asset ID
     /// @param amt The given amount
     function _give(
-        UserOrAccount memory userOrAccount,
+        uint256 userId,
         address receiver,
         uint256 assetId,
         uint128 amt
-    ) internal {
+    ) internal onlyAccountOwner(userId) {
```
xmxanuel, line 474: I prefer a design, which only uses modifier for `external` and `public` function.
The method is anyway internal, it is easier to understand if all access checks happening and the public function.

src/DripsHub.sol
```diff
@@ -931,32 +870,8 @@ abstract contract DripsHub {
...
-        }
-    }
-
     function calcUserId(address user) public pure returns (uint256) {
-        return uint256(keccak256(abi.encode(user)));
-    }
-
-    function calcUserId(address user, uint256 account) public pure returns (uint256) {
-        return uint256(keccak256(abi.encode(user, account)));
+        return uint160(user);
```
xmxanuel, line 874: I see this is also part of the intermediate code.

src/DripsHub.sol
```diff
@@ -789,32 +729,31 @@ abstract contract DripsHub {
     /// of the user or the account.
     /// If this is the first update, pass an empty array.
     function _assertCurrDrips(
-        UserOrAccount memory userOrAccount,
+        uint256 userId,
```
xmxanuel, line 732: No, UserOrAccount struct in memory anymore. :+1:

src/DripsHub.sol
```diff
@@ -242,6 +192,47 @@ abstract contract DripsHub {
...
+        }
+        _;
+    }
+
+    /// @notice Creates an account.
+    /// Assigns it an ID and lets its owner perform actions on behalf of all its sub-accounts.
+    /// Multiple accounts can be registered for a single address, it will own all of them.
+    /// @return accountId The new account ID.
+    function createAccount(address owner) public virtual returns (uint32 accountId) {
+        DripsHubStorage storage dripsHubStorage = _dripsHubStorage();
```
xmxanuel, line 217: I think we could shorten the naming a bit with the storage variables by changing the structs and helper method.

I would just call the dripsHubStorage `s` everywhere and maybe DripsHubStorage to SlotStorage.
The notation with `s` will be easily understood because you can see it from normally a line above.

The brain can understand such code quicker compared with long names
```solidity
 SlotStorage storage s = _storage();
 accountId = s.lastAccountId + 1;
 s.lastAccountId = accountId;
 s.accountsOwners[accountId] = owner;
```

xmxanuel (approved):
:+1: 

looks very good. I am an fan of the simple new userIds.

Only some ideas to improve code readability + I would like to move the modifier to the public functions.

src/DripsHub.sol
CodeSandwich: That's a good idea. It seems like we're going to divide the code into libs, so we're going to redo storage access anyway so we pass around the storage reference. Maybe we should apply this improvement then?

src/DripsHub.sol
CodeSandwich: It is, it's dropped in https://github.com/radicle-dev/radicle-drips-hub/pull/115.

src/DripsHub.sol
CodeSandwich: Me too, but take a look how this function is used, it's only a passthrough implementation of `give` in `ERC20DripsHub` and `EthDripsHub`. If we remove `onlyAccountOwner` here, we'd need to remember to add it in every implementation of `give`, that's poor separation of concerns and a good source of bugs.

PR #113, opened 2022-03-16
Receiver user IDs

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-drips-hub/issues/98, blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/112.

Switches the receiver configurations from addresses to user IDs for `setDrips`, `setSplits` and `give`. Now it's possible to support any sub-account, but only theoretically, there is still no way to collect for any sub-account, that'll be implemented in the upcoming PRs.

### DIFF
--- a/src/DaiDripsHub.sol
+++ b/src/DaiDripsHub.sol
@@ -84,7 +84,7 @@ contract DaiDripsHub is ERC20DripsHub {
     /// These parameters will be passed to the Dai contract by this function.
     /// @param permitArgs The Dai permission arguments.
     function giveAndPermit(
-        address receiver,
+        uint256 receiver,
         uint128 amt,
         PermitArgs calldata permitArgs
     ) public whenNotPaused {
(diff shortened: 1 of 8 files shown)


src/DaiDripsHub.sol
```diff
@@ -84,7 +84,7 @@ contract DaiDripsHub is ERC20DripsHub {
     /// These parameters will be passed to the Dai contract by this function.
     /// @param permitArgs The Dai permission arguments.
     function giveAndPermit(
-        address receiver,
+        uint256 receiver,
```
xmxanuel, line 87: One idea to increase the readability of the code can be to introduce a struct with only one uint256 as member for UserIds.

```solidity
struct UserId {
uint256 value;
}
```

The gas cost would be the same because in both cases it is just a `uint` just some syntax sugar for solidity, but it would allow 

```solidity
function functionName(UserId receiver,...)
```

xmxanuel (approved):
looks good

src/DaiDripsHub.sol
xmxanuel: Maybe even better, if we want to implement it. Could be still an overkill.

```
// struct
uint4 accountId;
uint28 subAccountId;
```

In that case we would avoid the bit manipulation. You could just access the `subAccountId`.

PR #115, opened 2022-03-16
Collectors user IDs

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/98, blocked by https://github.com/radicle-dev/radicle-drips-hub/pull/114.

Switches the collecting functions identification from implicit msg.sender to user ID. This means that sub-accounts can collect.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -121,12 +121,12 @@ abstract contract DripsHub {
     event Collectable(uint256 indexed userId, uint256 assetId, uint128 amt);
 
     /// @notice Emitted when drips are received and are ready to be split.
-    /// @param user The user
+    /// @param userId The user ID
     /// @param assetId The used asset ID
     /// @param amt The received amount.
     /// @param receivableCycles The number of cycles which still can be received.
     event ReceivedDrips(
-        address indexed user,
+        uint256 indexed userId,
         uint256 assetId,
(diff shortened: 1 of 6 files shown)


src/DripsHub.sol
```diff
@@ -94,37 +94,37 @@ abstract contract DripsHub {
     );
 
     /// @notice Emitted when the user's splits are updated.
-    /// @param user The user
+    /// @param userId The user ID
     /// @param receivers The list of the user's splits receivers.
-    event SplitsUpdated(uint256 indexed user, SplitsReceiver[] receivers);
+    event SplitsUpdated(uint256 indexed userId, SplitsReceiver[] receivers);
```
xmxanuel, line 99: Do you think we should have the `accountId` indexed as well?

I don't think so. We would say, use the events from the ERC721 if you want to see all tokenHolder drips?

Is this a correct assumption?

xmxanuel (approved):

src/DripsHub.sol
CodeSandwich: Yeah, I think so too. The DripsHub events should be agnostic about the meaning of the userId, it should be just a number. If you know how to interpret it, good for you, but it's not what DripsHub should focus on and make its users pay for.

PR #116, opened 2022-03-21
Drop special DripsHubs

CodeSandwich:
We don't plan to support Ether directly in DripsHub, so EthDripsHub is dead code. Until now it wasn't difficult to keep but we're about to add identity contracts and we'd need to provide an ether version as well, which would be more dead code just to support tests for existing dead code.

DaiDripsHub also doesn't make much sense in context of identity contracts, support for `permit` must be added to the identity API level. On DripsHub level `permit` would only accept messages signed by identity contracts, which they can't do because they don't have private keys.


xmxanuel (approved):
I agree. 
We could just use WETH and write some code to convert it before and after.

No, need to directly support it in the hub. It reduces the code complexitity.

PR #122, opened 2022-04-11
Add scheduled drips

CodeSandwich:
Adds scheduled drips. There are no new tests yet, let's see how this PR goes.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -4,6 +4,8 @@ pragma solidity ^0.8.7;
 struct DripsReceiver {
     uint256 userId;
     uint128 amtPerSec;
+    uint64 start;
+    uint64 duration;
 }
 
 library Drips {
@@ -84,6 +86,7 @@ library Drips {
     /// @notice Counts cycles from which drips can be collected.
     /// This function can be used to detect that there are
     /// too many cycles to analyze in a single transaction.
+    /// @param s The drips storage
     /// @param userId The user ID
     /// @param assetId The used asset ID
     /// @return cycles The number of cycles which can be flushed
@@ -100,6 +103,7 @@ library Drips {
     }
 
     /// @notice Calculate effects of calling `receiveDrips` with the given parameters.
+    /// @param s The drips storage
     /// @param userId The user ID
     /// @param assetId The used asset ID
     /// @param maxCycles The maximum number of received drips cycles.
@@ -131,6 +135,7 @@ library Drips {
     /// @notice Receive drips from uncollected cycles of the user.
     /// Received drips cycles won't need to be analyzed ever again.
     /// Calling this function does not collect but makes the funds ready to be split and collected.
+    /// @param s The drips storage
     /// @param userId The user ID
     /// @param assetId The used asset ID
     /// @param maxCycles The maximum number of received drips cycles.
@@ -168,6 +173,7 @@ library Drips {
     }
 
     /// @notice Current user drips hash, see `hashDrips`.
+    /// @param s The drips storage
     /// @param userId The user ID
     /// @param assetId The used asset ID
     /// @return currDripsHash The current user account's drips hash
@@ -179,24 +185,24 @@ library Drips {
         return s.dripsStates[userId][assetId].dripsHash;
     }
 
-    /// @notice Sets the user's or the account's drips configuration.
-    /// Transfers funds between the user's wallet and the drips hub contract
-    /// to fulfill the change of the drips balance.
+    /// @notice Sets the user's drips configuration.
+    /// @param s The drips storage
+    /// @param cycleSecs_ The cycle length
     /// @param userId The user ID
     /// @param assetId The used asset ID
-    /// @param lastUpdate The timestamp of the last drips update of the user or the account.
+    /// @param lastUpdate The timestamp of the last drips update of the user.
     /// If this is the first update, pass zero.
-    /// @param lastBalance The drips balance after the last drips update of the user or the account.
+    /// @param lastBalance The drips balance after the last drips update of the user.
     /// If this is the first update, pass zero.
     /// @param currReceivers The list of the drips receivers set in the last drips update
-    /// of the user or the account.
+    /// of the user.
     /// If this is the first update, pass an empty array.
-    /// @param balanceDelta The drips balance change to be applied.
-    /// Positive to add funds to the drips balance, negative to remove them.
-    /// @param newReceivers The list of the drips receivers of the user or the account to be set.
-    /// Must be sorted by the receivers' addresses, deduplicated and without 0 amtPerSecs.
-    /// @return newBalance The new drips balance of the user or the account.
-    /// Pass it as `lastBalance` when updating that user or the account for the next time.
+    /// @param balanceDelta The drips balance change being applied.
+    /// Positive when adding funds to the drips balance, negative to removing them.
+    /// @param newReceivers The list of the drips receivers of the user to be set.
+    /// Must be sorted, deduplicated and without 0 amtPerSecs.
+    /// @return newBalance The new drips balance of the user.
+    /// Pass it as `lastBalance` when updating that user for the next time.
     /// @return realBalanceDelta The actually applied drips balance change.
     function setDrips(
         Storage storage s,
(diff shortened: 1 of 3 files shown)


src/Drips.sol
```diff
@@ -358,14 +266,132 @@ library Drips {
...
+        uint256 length = 0;
+
+        for (uint256 i = 0; i < receivers.length; i++) {
+            DripsReceiver memory receiver = receivers[i];
+            require(receiver.amtPerSec != 0, "Drips receiver amtPerSec is zero");
+            require(i == 0 || _isOrdered(receivers[i - 1], receiver), "Receivers not sorted");
+            uint64 start = receiver.start;
+            if (start < dripsStart) start = dripsStart;
+            if (receiver.duration == 0) {
+                length = _addDefaultEndReceiver(defaults, length, start, receiver.amtPerSec);
```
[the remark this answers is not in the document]
xmxanuel: I think we can solve the ordering check very efficient.

As already said, we store the index of the matching index in the first array and remember the highest value from the first array ordering perspective to avoid double entries in the second array.

src/Drips.sol
```diff
@@ -358,14 +266,132 @@ library Drips {
...
+    /// @return newLength The new length of the list
+    function _addDefaultEndReceiver(
+        DefaultEndReceiver[] memory receivers,
+        uint256 length,
+        uint64 start,
+        uint128 amtPerSec
+    ) private pure returns (uint256 newLength) {
+        for (uint256 i = 0; i < length; i++) {
+            DefaultEndReceiver memory receiver = receivers[i];
+            if (receiver.start == start) {
```
xmxanuel, line 322: I think we should call the struct `DefaultEndTimes` because they are not really receivers because we missing the `address`

src/Drips.sol
```diff
@@ -358,14 +266,132 @@ library Drips {
...
+                receiver.amtPerSec += amtPerSec;
+                return length;
+            }
+            if (receiver.start > start) {
+                // Shift existing entries to make space for inserting the new entry
+                for (uint256 j = length; j > i; j--) {
```
xmxanuel, line 328:
```suggestion
                for (uint256 j = length+1; j > i; j--) {
```
I think, we first need to increase the length to make this algorithm work.

src/Drips.sol
```diff
@@ -377,113 +403,167 @@ library Drips {
...
+                    newDefaultEnd
                 );
-                newIdx++;
-            }
-            // The receiver may have never been used
-            if (!pickCurr) {
-                DripsState storage dripsState = s.dripsStates[receiver][assetId];
-                // The receiver has never been used, initialize it
+                _clearDeltaRange(deltas, cycleSecs, currStart, newStart, currRecv.amtPerSec);
+                _setDeltaRange(deltas, cycleSecs, currEnd, newEnd, currRecv.amtPerSec);
```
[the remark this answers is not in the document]
xmxanuel: Resolved. Already discussed in person.

xmxanuel (commented):
really good work. :rocket: 

Here my first comments. I still want to take a deeper look on certain aspects.

I added some ideas for different designs.
I think I found some minor issues, we would have discovered later with the tests.

src/Drips.sol
CodeSandwich: We've discussed it and it doesn't need fixing

src/Drips.sol
```diff
@@ -210,30 +216,22 @@ library Drips {
...
-                lastUpdate,
-                lastBalance,
-                currEndTime,
-                currAmtPerSec,
-                balanceDelta
-            );
-            newEndTime = _dripsEndTime(_currTimestamp(), newBalance, newAmtPerSec);
-        }
-        _updateDripsReceiversStates(
+        uint64 currDefaultEnd = _defaultEnd(lastBalance, lastUpdate, currReceivers);
```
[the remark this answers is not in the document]
CodeSandwich: We'll store it on-chain: https://github.com/radicle-dev/radicle-drips-hub/issues/128

src/Drips.sol
```diff
@@ -358,14 +266,132 @@ library Drips {
...
+    /// @param receivers The list of drips receivers.
+    /// @return balance The current drips balance.
+    function _currBalance(
+        uint128 startBalance,
+        uint64 lastUpdate,
+        uint64 defaultEnd,
+        DripsReceiver[] memory receivers
+    ) private view returns (uint128 balance) {
+        balance = startBalance;
+        for (uint256 i = 0; i < receivers.length; i++) {
```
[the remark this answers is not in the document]
CodeSandwich: Nice idea, but as long as we have receivers sorted by their user ID, don't we still need to iterate over all of them?

src/Drips.sol
xmxanuel, line 469: I think we should make the `_setDeltaRange` here twice. Using the `_clearDeltaRange` here is confusing.

src/Drips.sol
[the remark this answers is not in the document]
CodeSandwich: That's a good effort to remove the sorting part of the contract. I'm a little bit worried about the complexity of what you're proposing. From the API perspective it's a redundant parameter which is also difficult to handle by having weird and strict rules about how it should be built. The smart contract logic also expands to optionally support it with verification.

I think that for now we should make it simple and optimize it later, ideally without blowing up the API or the contract.

I've tried coming up with a solution not requiring receivers to be sorted by start time and failed. Finding the default end seems to be logically impossible without an ordered iteration. We've already discussed other API designs and while they wouldn't require sorting, they'd be much less useful.

I've tried different sorting approaches, but they usually boil down to linear searching, which with a hard limit of 100 items isn't unbearable. I've experimented with merge sort, which is theoretically more efficient, but for small lists it turned out to be much less efficient than the naiive algorithms.

src/Drips.sol
CodeSandwich: Well, `DefaultEndTime` doesn't contain the end time, so it's not great either. How about just `DefaultEnd`?

src/Drips.sol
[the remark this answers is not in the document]
xmxanuel: I think it make sense to get all PRs in. After we have all pieces together, we could still explore potential optimization.

From my perspective it is not such a weird way. We are giving a user the option to pass a specific sorted array instead of building it on-chain.

The verification can be done in O(n) and we anyway need to iterate the array.

However, i agree with the complexity perspective.  I think it is basically a trade-off between gas-optimisation and contract complexity.

I am fine with not implementing it right now. Let's try to get all PR's in. We can still discuss potential optimizations later.

xmxanuel (approved):

src/Drips.sol
CodeSandwich: But then we'd need to switch `currStart` and `newStart`, IMO that'd be weird and more difficult to visualize. If you have a strong opinion on that, I'm ok with changing that.

src/Drips.sol
CodeSandwich: Fixed

src/Drips.sol
```diff
@@ -358,14 +266,132 @@ library Drips {
...
+    }
+
+    /// @notice Calculates the current drips balance.
+    /// @param startBalance The balance when drips have started
+    /// @param lastUpdate The timestamp when drips have started.
+    /// @param defaultEnd The end time of drips without duration
+    /// @param receivers The list of drips receivers.
+    /// @return balance The current drips balance.
+    function _currBalance(
+        uint128 startBalance,
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

src/Drips.sol
xmxanuel: It depends on which layer you want to understand it. 
For me it caused confusion because i didn't understand why both start times are passed. 

Maybe at this level it would be more clear with something like this?

```solidity
// move start and end only if it has changed
moveDelta(currStart, currEnd, newStart, newEnd, currRecv.amtPerSec);
```
If it doesn't produce any additional gas.

CodeSandwich: Ok, that's good. I'm quite certain that it'll be inlined. Added.

PR #123, opened 2022-04-12
Drop DripsHub from asset flow

CodeSandwich:
This saves a `transfer` when moving tokens into or out of the drips contract.

### DIFF
--- a/src/AddressId.sol
+++ b/src/AddressId.sol
@@ -2,15 +2,18 @@
 pragma solidity ^0.8.7;
 
 import {DripsReceiver, ERC20DripsHub, SplitsReceiver} from "./ERC20DripsHub.sol";
+import {IERC20Reserve} from "./ERC20Reserve.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
 
 contract AddressId {
     ERC20DripsHub public immutable dripsHub;
+    address public immutable reserve;
     uint32 public immutable accountId;
 
     /// @param _dripsHub The drips hub to use
     constructor(ERC20DripsHub _dripsHub) {
         dripsHub = _dripsHub;
+        reserve = address(_dripsHub.reserve());
         accountId = _dripsHub.createAccount(address(this));
     }
 
@@ -118,8 +121,8 @@ contract AddressId {
 
     function _transferFromCaller(IERC20 erc20, uint128 amt) internal {
         require(erc20.transferFrom(msg.sender, address(this), amt), "Transfer from caller failed");
-        if (erc20.allowance(address(this), address(dripsHub)) < amt) {
-            erc20.approve(address(dripsHub), type(uint256).max);
+        if (erc20.allowance(address(this), reserve) < amt) {
+            erc20.approve(reserve, type(uint256).max);
         }
     }
 
--- a/src/ERC20Reserve.sol
+++ b/src/ERC20Reserve.sol
@@ -5,20 +5,28 @@ import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
 
 interface IERC20Reserve {
-    function withdraw(IERC20 token, uint256 amt) external;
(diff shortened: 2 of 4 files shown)


src/ERC20Reserve.sol
```diff
@@ -29,23 +37,35 @@ contract ERC20Reserve is IERC20Reserve, Ownable {
...
-    function deposit(IERC20 token, uint256 amt) public override onlyUser {
+    function deposit(
+        IERC20 token,
+        address from,
+        uint256 amt
+    ) public override onlyUser {
         withdrawable[token] += amt;
-        require(token.transferFrom(msg.sender, address(this), amt), "Reserve: transfer failed");
-        emit Deposited(msg.sender, token, amt);
+        require(token.transferFrom(from, address(this), amt), "Reserve: transfer failed");
```
xmxanuel, line 58: Why are we changing this?

For me it is important the user only will approve the DripsHub and not the Reserve.

Otherwise, the governance could change the reserve user and take all approved funds of a user from his/her EOA.

src/AddressId.sol
```diff
@@ -118,8 +121,8 @@ contract AddressId {
 
     function _transferFromCaller(IERC20 erc20, uint128 amt) internal {
         require(erc20.transferFrom(msg.sender, address(this), amt), "Transfer from caller failed");
-        if (erc20.allowance(address(this), address(dripsHub)) < amt) {
-            erc20.approve(address(dripsHub), type(uint256).max);
+        if (erc20.allowance(address(this), reserve) < amt) {
+            erc20.approve(reserve, type(uint256).max);
```
xmxanuel, line 125: This pattern together with removing the `msg.sender` from the reserve is a bit of a security issue.

It would be somehow dangerous to use the dripsHub as an EOA.

src/ERC20Reserve.sol
```diff
@@ -29,23 +37,35 @@ contract ERC20Reserve is IERC20Reserve, Ownable {
...
+        address to,
+        uint256 amt
+    ) public override onlyUser {
         uint256 balance = withdrawable[token];
         require(balance >= amt, "Reserve: withdrawal over balance");
         withdrawable[token] = balance - amt;
-        require(token.transfer(msg.sender, amt), "Reserve: transfer failed");
-        emit Withdrawn(msg.sender, token, amt);
+        require(token.transfer(to, amt), "Reserve: transfer failed");
+        emit Withdrawn(msg.sender, token, to, amt);
```
xmxanuel, line 49:
```suggestion
        emit Withdraw(msg.sender, token, to, amt);
```

xmxanuel: I think present tense is nicer for the event names

src/ERC20Reserve.sol
CodeSandwich: The actual user (the EOA signer) needs to approve the `AddressId` contract anyway, not the `DripsHub`. This PR changes the address to approve by the identity contract, which only has an impact on the developers of such contracts.

This PR doesn't change the trust assumptions toward the governance. You're right about the possible attack vector in case the governance goes rogue, but in current setup it can just upgrade DripsHub to abuse all the approvals.

src/ERC20Reserve.sol
CodeSandwich: FWIW OpenZeppelin consistently uses past tense. ERCs sometimes use past tenses as well (e.g. 777) or nouns derived from verbs (e.g. 20, 721, 1155). IMO if we want to use verbs, we probably should stick to the past forms to lower the surprise factor :shrug:

src/AddressId.sol
CodeSandwich: Could you elaborate about the security impact? I don't understand the `use the dripsHub as an EOA` part.

xmxanuel: If user want to generate subAccounts with an EOA address for whatever reason.
In that case If I understood correctly an approval on the reserve would be required.

However, I agree we have the same problem with Governance upgradability.

xmxanuel (approved):

PR #124, opened 2022-04-13
Make ManagedDripsHub a mix-in

CodeSandwich:
Blocked by #123.

The hierarchy is changed from `ERC20DripsHub -> ManagedDripsHub -> DripsHub` to `ERC20DripsHub -> (Managed, DripsHub)`. `Managed` is now generic and not tied to `DripsHub`, this improves separation of concerns and enables flattening the abstract `DripsHub` and `ERC20DripsHub` into a single contract, which in turn allows its API to be better tailored for ERC-20 tokens.


src/Managed.sol
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// @notice Changes the admin of the contract.
+    /// Can only be called by the current admin.
+    function changeAdmin(address newAdmin) public onlyAdmin {
+        _changeAdmin(newAdmin);
+    }
+
+    /// @notice Throws if called by any account other than the admin.
+    modifier onlyAdmin() {
```
xmxanuel, line 64: Maybe we can move the `modifier` definitions after the constructor before the regular functions.

xmxanuel (approved):
This is a very good idea :+1:

src/Managed.sol
CodeSandwich: Fixed

PR #125, opened 2022-04-13
Flatten ERC20DripsHub into DripsHub

CodeSandwich:
Blocked by #124.

Makes DripsHub ERC-20-only. This simplifies the contracts and allows the API to be better tailored for ERC-20 assets.


src/DripsHub.sol
```diff
@@ -274,7 +275,7 @@ abstract contract DripsHub {
         uint256 userId,
         uint256 assetId,
         SplitsReceiver[] memory currReceivers
-    ) public virtual returns (uint128 collectableAmt, uint128 splitAmt) {
+    ) public whenNotPaused returns (uint128 collectableAmt, uint128 splitAmt) {
```
xmxanuel, line 278: Now, it is way cleaner. Just the modifier here without the wrapper functions :+1:

src/DripsHub.sol
```diff
@@ -427,5 +434,23 @@ abstract contract DripsHub {
     /// @param assetId The used asset ID
     /// @param amt The transferred amount.
     /// Positive to transfer funds to the user, negative to transfer from them.
-    function _transfer(uint256 assetId, int128 amt) internal virtual;
+    function _transfer(uint256 assetId, int128 amt) internal {
+        IERC20 erc20 = IERC20(address(uint160(assetId)));
+        if (amt > 0) {
+            reserve.withdraw(erc20, msg.sender, uint128(amt));
+        } else if (amt < 0) {
+            reserve.deposit(erc20, msg.sender, uint128(-amt));
```
xmxanuel, line 442: I would still prefer to make it vice-versa to be aligned with the `balanceDelta`.

```solidity
/// Positive to add funds to the drips balance, negative to remove them.
```

xmxanuel (approved):

src/DripsHub.sol
CodeSandwich: Agree, fixed

PR #126, opened 2022-04-13
Accept IERC20 in DripsHub API

CodeSandwich:
Blocked by #125 

Switches DripsHub API from requiring leakingly abstract assetIDs to plain IERC20 addresses.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -151,15 +151,16 @@ contract DripsHub is Managed {
 
     /// @notice Returns amount of received funds available for collection for a user.
     /// @param userId The user ID
-    /// @param assetId The used asset ID
+    /// @param erc20 The used ERC-20 token
     /// @param currReceivers The list of the user's current splits receivers.
     /// @return collectedAmt The collected amount
     /// @return splitAmt The amount split to the user's splits receivers
     function collectableAll(
         uint256 userId,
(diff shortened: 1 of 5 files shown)


src/DripsHub.sol
```diff
@@ -453,4 +453,11 @@ contract DripsHub is Managed {
             storageRef.slot := slot
         }
     }
+
+    /// @notice Generates an asset ID for the ERC-20 token
+    /// @param erc20 The ERC-20 token
+    /// @return assetId The asset ID
+    function _assetId(IERC20 erc20) internal pure returns (uint256 assetId) {
+        return uint160(address(erc20));
+    }
```
xmxanuel, line 462: I still think this might be a bit over-engineered or I forgot why we are doing it. 

Why are we not just using an `address` as asset type?

xmxanuel (approved):

src/DripsHub.sol
CodeSandwich: We're doing it to make the Drips library as generic and unconstrained as possible. The library doesn't care what the assetId represents, we're using it for ERC-20 addresses, but others could use it to encode any kinds of things, e.g. they may want to encode drips in Ether, ERC-1155s or even some custom asset IDs. From the performance perspective it doesn't matter as long as it fits in a single word.

Issue #101, opened 2022-02-15
add optional reserve plugins


xmxanuel:
- like DSR saving rates

deckardkrug:
There are two patterns I have thought of.

**1.** We use `inheritance` to override `fn deposit` and `fn withdraw` to enable the interaction with another contract to use e.g. DSR saving rates.

Pros:
- clean logic

Cons:
- enforces multiple deployments for different reserves with different plugins
- front-end integration can be very hard as we would need to keep a registry for all the different contracts
- user would need to move his assets between multiple reserve contracts just in order to use another plugin

---

**2.** We use `delegatecall` to extend or override the deposit and withdraw functions. In this case we would add a mapping that stores a user address and the current plugin address that the user added and `fn switchPlugin` which updates the mentioned mapping and ensures that the user does not lose access to his assets if he wants to change the plugin.

There are two variants of this implementation:
-  a wrapper that implements the mentioned mapping and functions
- implement the mentioned mapping and functions in the current reserve contract

Pros:
- more flexibility
- only one deployment
- user could pass any contract that supports a plugin interface
- every user could have a different plugin without changing the the reserve addresses

Cons:
- potential security risk if `delegatecall` is implemented wrong

xmxanuel:
@deckardkrug good ideas 👍 

who do you mean with the term `user` in that case?

- a user of the reserve (in our case the dripsHub contract) 
- a user of the DripsHub

Our idea was the the governance would decide for all Dai in the reserve.  The complexity might increase if we start to do it per token type. 

Currently, no  drips user would need to interact with the reserve.  We route the Dai through the dripsHub, so there should be no UI implications. 
https://github.com/radicle-dev/radicle-drips-hub/blob/master/src/ERC20DripsHub.sol#L159

In case we would implement multiple different plugins/adapters a contract using the reserve interface would perform the orchestration. At least this was the idea behind it.

Let say the governance decides USDC should be deposited into Uniswap. 

We would have some kind of register method 
```solidity
registerPlugin(address plugin, address er20) public onlyGovernance
```

We would implement small uniswap adapter contract which is afterwards registered. Every `deposit` or `withdraw` would be forwarded to the adapter.

xmxanuel:
@CodeSandwich  just saw your comment my `registerPlugin` is basically your `setPlugin` from the doc. :)

CodeSandwich:
[...]

PR #129, opened 2022-05-03
Add reserve plugins

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/101.

### DIFF
--- a/src/Reserve.sol
+++ b/src/Reserve.sol
@@ -4,30 +4,134 @@ pragma solidity ^0.8.7;
 import {Ownable} from "openzeppelin-contracts/access/Ownable.sol";
 import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
 
+/// @notice The reserve interface as seen by the users.
 interface IReserve {
-    function withdraw(
-        IERC20 token,
-        address to,
-        uint256 amt
-    ) external;
-
+    /// @notice Deposits funds into the reserve.
+    /// The reserve will `transferFrom` `amt` tokens from the `from` address.
+    /// @param token The used token.
+    /// @param from The address from which funds are deposited.
+    /// @param amt The deposited amount.
     function deposit(
         IERC20 token,
         address from,
         uint256 amt
     ) external;
+
+    /// @notice Withdraws funds from the reserve.
+    /// The reserve will transfer `amt` tokens to the `to` address.
+    /// Only funds previously deposited can be withdrawn.
+    /// @param token The used token.
+    /// @param to The address to which funds are withdrawn.
+    /// @param amt The withdrawn amount.
+    function withdraw(
+        IERC20 token,
+        address to,
+        uint256 amt
+    ) external;
 }
 
-contract Reserve is IReserve, Ownable {
-    mapping(address => bool) public isUser;
-    mapping(IERC20 => uint256) public withdrawable;
+/// @notice The reserve plugin interface required by the reserve.
+interface IReservePlugin {
+    /// @notice Called by the reserve when it starts using the plugin,
+    /// immediately after transferring to the plugin all the deposited funds.
+    /// This initial transfer won't trigger the regular call to `afterDeposition`.
+    /// @param token The used token.
+    /// @param amt The amount which has been transferred for deposition.
+    function afterStart(IERC20 token, uint256 amt) external;
 
-    event Withdrawn(address user, IERC20 indexed token, address indexed to, uint256 amt);
+    /// @notice Called by the reserve immediately after
+    /// transferring funds to the plugin for deposition.
+    /// @param token The used token.
+    /// @param amt The amount which has been transferred for deposition.
+    function afterDeposition(IERC20 token, uint256 amt) external;
+
+    /// @notice Called by the reserve right before transferring funds for withdrawal.
+    /// The reserve will `transferFrom` the tokens from the plugin address.
+    /// The reserve can always withdraw everything that has been ever deposited, but never more.
+    /// @param token The used token.
+    /// @param amt The amount which will be transferred.
+    function beforeWithdrawal(IERC20 token, uint256 amt) external;
+
+    /// @notice Called by the reserve when it stops using the plugin,
+    /// right before transferring from the plugin all the deposited funds.
+    /// The reserve will `transferFrom` the tokens from the plugin address.
+    /// This final transfer won't trigger the regular call to `beforeWithdrawal`.
+    /// @param token The used token.
+    /// @param amt The amount which will be transferred.
+    function beforeEnd(IERC20 token, uint256 amt) external;
+}
+
+/// @notice The ERC-20 tokens reserve contract.
+/// The registered users can deposit and withdraw funds.
+/// The reserve by default doesn't do anything with the tokens,
+/// but for each ERC-20 address a plugin can be registered for tokens storage.
+contract Reserve is IReserve, Ownable {
+    /// @notice The dummy plugin address meaning that no plugin is being used.
+    IReservePlugin public constant NO_PLUGIN = IReservePlugin(address(0));
+
+    /// @notice A set of addresses considered users.
+    /// The value is `true` if an address is a user, `false` otherwise.
+    mapping(address => bool) public isUser;
+    /// @notice How many tokens are deposited for each token address.
+    mapping(IERC20 => uint256) public deposited;
+    /// @notice The reserved plugins for each token address.
+    mapping(IERC20 => IReservePlugin) public plugins;
+
+    /// @notice Emitted when a plugin is set.
+    /// @param owner The address which called the function.
+    /// @param token The token for which plugin has been set.
+    /// @param oldPlugin The old plugin address. `NO_PLUGIN` if no plugin was being used.
+    /// @param newPlugin The new plugin address. `NO_PLUGIN` if no plugin will be used.
+    /// @param amt The amount which has been withdrawn
+    /// from the old plugin and deposited into the new one.
+    event PluginSet(
+        address owner,
+        IERC20 indexed token,
+        IReservePlugin indexed oldPlugin,
+        IReservePlugin indexed newPlugin,
+        uint256 amt
+    );
+
+    /// @notice Emitted when funds are deposited.
+    /// @param user The address which called the function.
+    /// @param token The used token.
+    /// @param from The address from which tokens have been transferred.
+    /// @param amt The amount which has been deposited.
     event Deposited(address user, IERC20 indexed token, address indexed from, uint256 amt);
-    event ForceWithdrawn(address owner, IERC20 indexed token, address indexed to, uint256 amt);
+
+    /// @notice Emitted when funds are withdrawn.
+    /// @param user The address which called the function.
+    /// @param token The used token.
+    /// @param to The address to which tokens have been transferred.
+    /// @param amt The amount which has been withdrawn.
+    event Withdrawn(address user, IERC20 indexed token, address indexed to, uint256 amt);
+
+    /// @notice Emitted when funds are force withdrawn.
+    /// @param owner The address which called the function.
+    /// @param token The used token.
+    /// @param plugin The address of the plugin from which funds have been withdrawn or
+    /// `NO_PLUGIN` if from the reserve itself.
+    /// @param to The address to which tokens have been transferred.
+    /// @param amt The amount which has been withdrawn.
+    event ForceWithdrawn(
+        address owner,
+        IERC20 indexed token,
+        IReservePlugin indexed plugin,
+        address indexed to,
+        uint256 amt
+    );
+
+    /// @notice Emitted when an address is registered as a user.
+    /// @param owner The address which called the function.
+    /// @param user The registered user address.
     event UserAdded(address owner, address indexed user);
+
+    /// @notice Emitted when an address is unregistered as a user.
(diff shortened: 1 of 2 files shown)


src/Reserve.sol
```diff
@@ -4,30 +4,115 @@ pragma solidity ^0.8.7;
...
+/// The registered users can deposit and withdraw funds.
+/// The reserve by default doesn't do anything with the tokens,
+/// but for each ERC-20 address a plugin can be registered for tokens storage.
 contract Reserve is IReserve, Ownable {
+    /// @notice The dummy plugin address meaning that no plugin is being used.
+    IReservePlugin public constant NO_PLUGIN = IReservePlugin(address(0));
+
+    /// @notice A set of addresses considered users.
+    /// The value is `true` if an address is a user, `false` otherwise.
     mapping(address => bool) public isUser;
```
xmxanuel, line 58: I think i would prefer `access`/ `allowance` here because the we use the term `user` already in the DripsHub context and ReserveUsers are other Drips smart contracts and not the Drips user

src/Reserve.sol
```diff
@@ -37,44 +122,146 @@ contract Reserve is IReserve, Ownable {
...
+    /// set for that token and deposited into the new one.
+    /// If no plugin has been set, funds are deposited from the reserve itself.
+    /// If no plugin is being set, funds are deposited into the reserve itself.
+    /// Callable only by the current owner.
+    /// @param token The used token.
+    /// @param newPlugin The new plugin address. `NO_PLUGIN` if no plugin should be used.
+    function setPlugin(IERC20 token, IReservePlugin newPlugin) public onlyOwner {
+        IReservePlugin oldPlugin = plugins[token];
+        plugins[token] = newPlugin;
+        uint256 amt = deposited[token];
```
xmxanuel, line 137: I think in the case of a plugin migration we should not use `deposited` amount.

Instead, we should take all the balance (deposit + interest/earned amount). Otherwise the interest gets lost.

I think each plugin should have a common interface for it.
```solidity
plugin.balance() // or totalAvialable
```

src/Reserve.sol
```diff
@@ -37,44 +122,146 @@ contract Reserve is IReserve, Ownable {
...
+        _beforeWithdrawal(token, oldPlugin, amt);
+        _transfer(token, _pluginAddr(oldPlugin), _pluginAddr(newPlugin), amt);
+        _afterDeposition(token, newPlugin, amt);
+        emit PluginSet(msg.sender, token, oldPlugin, newPlugin);
     }
 
+    /// @notice Deposits funds into the reserve.
```
xmxanuel, line 144: It would be nice to add the following method.

```solidity
function surplus(IERC20 erc20) external view returns (int surplus) {
 IReservePlugin plugin = plugins[token];
// could be negative as well
 returns plugin.balance() - deposited[token];
}

```

Currently, we don't have a strategy what should happen with a `surplus` but I think for now we should assume to governance should be able to take the `surplus` with a governance call forced `withdraw`

xmxanuel: However, it might be too much work to start a governance process each time.

Instead i could imagine the following. The `onlyOwner` can set a `surplusReceiver` for a specific token.

Together with a public callable `surplusHarvest` method, which pays out the difference between `plugin.balance() - deposited[token]` if `surplusReceiver` is set.

src/Reserve.sol
```diff
@@ -37,44 +122,146 @@ contract Reserve is IReserve, Ownable {
...
+    /// @notice Sets the deposited amount counter for a token without transferring any funds.
+    /// If used recklessly, it may cause a mismatch between the counter and the actual balance
+    /// making valid future calls to `withdraw` or `setPlugin` fail due to lack of funds.
+    /// It may also make the counter lower than what users expect it to be again making
+    /// valid future calls to `withdraw` fail.
+    /// Callable only by the current owner.
+    /// @param token The used token.
+    /// @param amt The new deposited amount counter value.
+    function setDeposited(IERC20 token, uint256 amt) public onlyOwner {
+        deposited[token] = amt;
```
xmxanuel, line 216: Do we really need this function? If we implement some surplus harvest pattern?

src/Reserve.sol
```diff
@@ -37,44 +122,146 @@ contract Reserve is IReserve, Ownable {
...
+        uint256 amt
+    ) internal {
+        bool success;
+        if (from == address(this)) {
+            success = token.transfer(to, amt);
+        } else {
+            success = token.transferFrom(from, to, amt);
+        }
+        require(success, "Reserve: transfer failed");
+    }
```
xmxanuel, line 266: We have already the global view function for `deposited(IERC20 token)` it would be nice to add one for the `balance` as well.

Basically a wrapper function which asks the balance of the plugin. In case of no plugin it might be `erc20.blanceOf(this)`.

src/Reserve.sol
```diff
@@ -4,30 +4,115 @@ pragma solidity ^0.8.7;
...
+    /// @param amt The amount which will be transferred.
+    function beforeWithdrawal(IERC20 token, uint256 amt) external;
+}
+
+/// @notice The ERC-20 tokens reserve contract.
+/// The registered users can deposit and withdraw funds.
+/// The reserve by default doesn't do anything with the tokens,
+/// but for each ERC-20 address a plugin can be registered for tokens storage.
 contract Reserve is IReserve, Ownable {
+    /// @notice The dummy plugin address meaning that no plugin is being used.
```
xmxanuel, line 53: I am wondering if we are completely open to support any ERC20 or if governance should have a potential `blackList` mapping.

If we want such a blacklist we potentially could manage it here directly. 

`deposit` and `withdraw` would just revert, if a token is on a blacklist.

src/Reserve.sol
xmxanuel: If we implement the `surplusHarvest` pattern. See other comment. We should call the `surplusHarvest` method before migrating to another reserve.

src/Reserve.sol
```diff
@@ -37,44 +122,146 @@ contract Reserve is IReserve, Ownable {
...
+    /// @param amt The deposited amount.
     function deposit(
         IERC20 token,
         address from,
         uint256 amt
     ) public override onlyUser {
-        withdrawable[token] += amt;
-        require(token.transferFrom(from, address(this), amt), "Reserve: transfer failed");
+        IReservePlugin plugin = plugins[token];
+        require(from != address(plugin) && from != address(this), "Reserve: deposition from self");
```
xmxanuel, line 156: `from` reserve would be a wild attack, good check. :)

src/Reserve.sol
```diff
@@ -4,30 +4,115 @@ pragma solidity ^0.8.7;
...
+    /// @notice Emitted when a plugin is set.
+    /// @param owner The address which called the function.
+    /// @param token The token for which plugin has been set.
+    /// @param oldPlugin The old plugin address. `NO_PLUGIN` if no plugin was being used.
+    /// @param newPlugin The new plugin address. `NO_PLUGIN` if no plugin will be used.
+    event PluginSet(
+        address owner,
+        IERC20 indexed token,
+        IReservePlugin indexed oldPlugin,
+        IReservePlugin indexed newPlugin
```
xmxanuel, line 73: I think it would the nice to add the `amount` we took from old plugin and moved into the new plugin to the event.

xmxanuel (approved):
Nice work. I like the design this gives us a lot of flexibility.

I added some ideas how to deal on a higher level with the `surplus` of a plugin and some minor improvements.

src/Reserve.sol
xmxanuel: However, no strong opinion.

CodeSandwich: I too am not a fan of `user`. Maybe `controller`?

xmxanuel: Maybe `Depositor` ?

src/Reserve.sol
CodeSandwich: I think that we should just have a `BlacklistPlugin`, which implements the plugin API, but reverts for whichever token it's assigned. It's much cleaner and more powerful, because it makes it possible to e.g. allow refunding of tokens already in the system.

src/Reserve.sol
CodeSandwich: Yes, we still do, there needs to be an emergency tool so if something goes awfully wrong, we can manually fix the state.

src/Reserve.sol
xmxanuel: Thats, a good idea. :+1:

src/Reserve.sol
xmxanuel: Yes, i can see the concern. Lets keep it, just to have a emergency mechanism.  The governance anyway has access to the funds as well.

src/Reserve.sol
CodeSandwich: That made me thinking that maybe the plugin API should implement a lifecycle for plugins making it easy to gracefully set them up and then shut down :thinking: E.g. implementation of the `BlacklistPlugin` would become trivial with such interface. 

```solidity
interface IReservePlugin {
    // Unchanged
    function afterDeposition(IERC20 token, uint256 amt) external;
    function beforeWithdrawal(IERC20 token, uint256 amt) external;

    // Acts like `afterDeposition`, but called when setting a plugin after the initial deposition
    function afterSetting(IERC20 token, uint256 amt) external;

    // Acts like `beforeWithdrawal`, but called before unsetting and draining the plugin
    function beforeUnsetting(IERC20 token, uint256 amt) external;
}
```

src/Reserve.sol
CodeSandwich: I'd be surprised to see `DripsHub` attacking the reserve, but it's still better to not have a built-in "feature" of doubling the balance and breaking the API :laughing:

src/Reserve.sol
CodeSandwich: I think that `balance()` is a good idea if we add the surplus support to `Reserve`. Otherwise it's not very useful, because it doesn't facilitate the usage of the `Reserve` API. It also adds a function to the plugin API which by its design makes an assumption about the balance is a single, readily available number. It also implies that this exact amount to be immediately withdrawable, which may not be true, because funds could be available only in the future.

src/Reserve.sol
xmxanuel: Yes, we have very high uncertainty in the design phase. Therefore, I agree to keep it as generic as possible.

>  Who can claim the surplus?

This was my idea with the setting a `surplusReceiver`.  I can see you reasoning. I think to keep the `surplus` completely out of the system also works.

We only need to be aware of the limitations. That a `plugin` migration would not migrate the surplus. The plugin needs to take care of it.

I think most likely, we will end up with common surplus strategy for all plugins or ERC20.

For example the entire surplus should go to this Multi-Sig or maybe later you want to change it that a specific reward contract should receive it.

It might be beneficial to change them one in all step. Which we might loss here. 

However, I don't see it as that critical because most likely we won't have that many.

CodeSandwich: > a `plugin` migration would not migrate the surplus

Yes, that's a drawback, it makes the plugin migration a more complex operation. We must decide what to do with the surplus and if we choose to move it to the new plugin, then we must organize it outside of the `Reserve`.

> It might be beneficial to change them one in all step.

Nothing is stopping us from registering a single smart contract for multiple ERC-20 tokens, we can then do all kinds of batch changes. We can also modularize further and have multiple plugins but all implementing a common surplus withdrawal API callable by a contract deciding what to do with these funds.

src/Reserve.sol
xmxanuel: Haha. I depends how you see it. You are seeing the entire system.  I see it only from the perspective of the Reserve. There is no DripsHub from the perspective of the Reserve.

Just smart contracts or EOA which can deposit funds.  Now, we need a good name for those addresses.

Here is my name canditate:
`depositor` =>  `a person who keeps money at a bank`. 
(See: https://dictionary.cambridge.org/de/worterbuch/englisch/depositor)

Which might be a good name candidate for your reserve user.

src/Reserve.sol
CodeSandwich: Ok, so let's keep this PR's design.

src/Reserve.sol
CodeSandwich: Added

src/Reserve.sol
xmxanuel: Yes, that's nicer. Otherwise, `afterDeposition` would have to handle the first deposit as an edge case.

Normally, the first step might require some initial steps.

```solidity
    function start(IERC20 token, uint256 amt) external;
    function shutdown(IERC20 token, uint256 amt) external;
}
```

I think from the naming the plugin is not aware that it is `set`. It can `start` and `shutdown` with a specific balance.

You get my naming philosophy. :)

xmxanuel: The lower layer naming should not reflect the higher layer specific purpose.

CodeSandwich: Added

src/Reserve.sol
xmxanuel: > It also adds a function to the plugin API which by its design makes an assumption about the balance being a single, readily available number.

Yes, but i think this a requirement. If some Plugin can not offer that, it might be not possible to withdraw at any point in time. Which we definitely need as a functionality.

I think it might be later useful to query certain information centralized.

It was just an idea.

PR #130, opened 2022-05-04
Upgrades

CodeSandwich:
Bump Solidity to 0.8.13 and upgrade ds-test and OpenZeppelin.


.dapprc
```diff
@@ -1,4 +1,4 @@
-export DAPP_SOLC_VERSION=0.8.7
```
xmxanuel, line 1: :clap:

xmxanuel (approved):
nice.

PR #131, opened 2022-05-06
Extract drips test

CodeSandwich:


xmxanuel (approved):
great work. Good idea to re-implement the helpers that the test can be easily migrated.

I think it make sense to test the library isolated.

I noticed some renaming of the test function names like `testDripsSomeFundsToTwoReceivers` to `testDripsToTwoReceivers`  and a couple more.

This makes it a bit hard to follow.

Did you migrate all tests? I think some of them are still missing correct?

CodeSandwich:
Sorry about the renames :grimacing: At least the order is the same. I've copied all the tests which could be applied on the lib itself and removed them from DripsHub where they were too low level to be kept.

Issue #84, opened 2022-01-19
Add scheduled drips


CodeSandwich:
It **may** be possible to support drips configurations with optional start and end time. The user would need to have balance covering at least all the drips with end times. Drips without end time probably should run for as long as balance allows.

Partially done in https://github.com/radicle-dev/radicle-drips-hub/issues/84, still to do:
- restore events emitted when setting drips
- add tests for the scheduled drips

deckardkrug:
Is someone already working on this?

CodeSandwich:
Currently it's just an idea, we need to take a closer look if it's even realistically doable. In short term we have more burning features to add, but this could be a valuable feature allowing more tools to be based on drips hub. One of the reasons why I've opened this issue is to signal a potential possibility and maybe get some feedback.

deckardkrug:
@CodeSandwich okay, thank you.

As I imagine the implementation of this idea, it could be a check in `setDrips()` and two additional parameters. What I like about this idea is not necessary the end time but the start time. Using a start time, streams could be "scheduled".

bordumb:
This would be a REALLY awesome feature.

I can even think of some IRL use cases:
* Time-limited fundraiser for an event: imagine you want to make a conference for 1,000 attendees. You could do a one-time fundraiser that starts at `day-0` and ends up `day-90`. Anyone who buys the NFT has a "ticket" for entry
* Time-limited charity auction: imagine you are raising money for some charity/cause. You could do a one-time fundraiser that starts at `day-0` and ends up `day-90`. Set limit to 10,000 NFTs. The NFT could become a raffle ticket, so maybe 1 out of the 10,000 will get some special prize and the rest of the funds go to a charity. **This would not be possible without a time limit**.

CodeSandwich:
> Question: is this JUST for the one-off Drips? Or could it implemented for Community building, too?

It'd be for the drips, which is the feature where funds are flowing at a constant rate over a period of time. I think that we mean the same thing :wink:  It also covers one-offs because they can be easily emulated with this. You've invented some awesome use cases!

bordumb:
FYI, I've continued my ideas on a new issue below
I think I was conflating **gives** and **drips** above
So the issue below focuses on **gives** within the context of fundraising 

https://github.com/radicle-dev/radicle-drips-hub/issues/89

Issue #86, opened 2022-01-26
Emit an event for the splits receiver when splits configured


CodeSandwich:
As of now when setting splits, a `SplitsUpdated` event is emitted which contains an unindexed list of receivers. Without an aggregator like a subgraph it's impossible for the receiver to detect this event and start monitoring the splitting user. Only if they decide to collect a `Split` event is emitted which can be noticed by the receiver. Until then there could be a hidden fortune of assets waiting to be split if only the receiver knew about its existence.

It would probably make sense to introduce a new event type:
```solidity
event Splitting(address indexed user, address indexed receiver, uint32 weight);
```

A series of events like this may or may not be enough to take over the function of a `SplitsUpdated` event. If that's the case, `SplitsUpdated` could be removed to save gas. If not, `user` probably doesn't need to be indexed.

PR #136, opened 2022-05-13
Redesign events around receivers lists

CodeSandwich:
Closes https://github.com/radicle-dev/radicle-drips-hub/issues/84 and https://github.com/radicle-dev/radicle-drips-hub/issues/86.

CC: @gh0stwheel

In this PR the `*ReceiverSeen` events aren't emitted when the user doesn't update their configuration to save gas. This can be further improved by maintaining a global set of already emitted receiver lists. The problem is that the storage cost is high but gas savings aren't guaranteed. If we go in that direction, events handling will also become easier, because there won't be a need to handle multiple copies of the same entries.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -1,10 +1,17 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.8.13;
 
+/// @notice A drips receiver
 struct DripsReceiver {
+    /// @notice The user ID.
     uint256 userId;
+    /// @notice The amount per second being dripped. Must never be zero.
     uint128 amtPerSec;
+    /// @notice The timestamp when dripping should start.
+    /// If zero, use the timestamp when drips are configured.
     uint64 start;
+    /// @notice The duration of dripping.
+    /// If zero, drip until balance runs out.
     uint64 duration;
 }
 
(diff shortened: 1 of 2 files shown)


src/Drips.sol
```diff
@@ -15,34 +22,32 @@ library Drips {
...
+    /// If zero, drip until balance runs out.
+    event DripsReceiverSeen(
+        bytes32 indexed receiversHash,
         uint256 indexed userId,
-        uint256 indexed assetId,
-        uint128 balance,
-        DripsReceiver[] receivers
+        uint128 amtPerSec,
+        uint64 start,
+        uint64 duration
```
xmxanuel, line 50: I am wondering if it makes sense to emit the `endTime` as well?

Obviously, you can calculate it based on the block.timestamp but we have some edge-cases like

`start < block.timestamp`  then the end is not `start + duration`

src/Drips.sol
```diff
@@ -15,34 +22,32 @@ library Drips {
...
-    /// @param assetId The used asset ID
-    /// @param balance The new drips balance. These funds will be dripped to the receivers.
-    /// @param receivers The new list of the drips receivers.
-    event DripsUpdated(
+    /// @param amtPerSec The amount per second being dripped. Must never be zero.
+    /// @param start The timestamp when dripping should start.
+    /// If zero, use the timestamp when drips are configured.
+    /// @param duration The duration of dripping.
+    /// If zero, drip until balance runs out.
+    event DripsReceiverSeen(
```
xmxanuel, line 45: Why not just `DripsReceiver`?

src/Splits.sol
```diff
@@ -205,8 +218,16 @@ library Splits {
         SplitsReceiver[] memory receivers
     ) internal {
         assertSplitsValid(receivers);
-        s.splitsStates[userId].splitsHash = hashSplits(receivers);
-        emit SplitsUpdated(userId, receivers);
+        SplitsState storage state = s.splitsStates[userId];
+        bytes32 newSplitsHash = hashSplits(receivers);
+        if (newSplitsHash != state.splitsHash) {
+            state.splitsHash = newSplitsHash;
+            for (uint256 i = 0; i < receivers.length; i++) {
```
xmxanuel, line 225: Here it is the same we iterate them already in the `assertSplitsValid` function. Why not just emit the events there?

src/Drips.sol
CodeSandwich: Addition of `endTime` would enforce re-emission of all the `DripsReceiverSeen` events in case of a top-up. Some off-chain complexity is IMO a good trade-off for potential gas savings, each `DripsReceiverSeen` costs about 1900 gas to emit.

src/Drips.sol
CodeSandwich: It would be nice, but won't it be confusing? Events usually describe things that happened, `DripsReceiver` doesn't happen. If you think it's ok, I'll be glad to shorten the name.

src/Drips.sol
```diff
@@ -251,10 +257,23 @@ library Drips {
...
+            state.dripsHash = newDripsHash;
+            for (uint256 i = 0; i < newReceivers.length; i++) {
+                DripsReceiver memory receiver = newReceivers[i];
+                emit DripsReceiverSeen(
+                    newDripsHash,
+                    receiver.userId,
+                    receiver.amtPerSec,
+                    receiver.start,
+                    receiver.duration
+                );
```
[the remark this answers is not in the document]
CodeSandwich: Because it would be quite cumbersome to do there, this function already does a lot, has a complex flow and can barely handle all its local variables. If the cost of the additional iteration isn't too high, I'd prefer keeping it in a separate loop and close to the hash updating logic for the sake of clarity.

src/Splits.sol
CodeSandwich: Done, also we don't validate the configuration anymore if it's unchanged.

src/Drips.sol
xmxanuel: We can keep it. No strong opinion

src/Drips.sol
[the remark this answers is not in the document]
xmxanuel: It is always the trade-off between gas costs and code complexity. I remember that it is not that expensive to iterate. Let's keep it for now.

xmxanuel (approved):

gh0stwheel:
Sorry it took me so long to respond to this.

It looks okay to me!

PR #137, opened 2022-05-23
test: migrate splits tests to only test the library

xmxanuel:
I migrated the splits tests from `DripsHub.t.sol` which only tests the Split library.

I modified some of them and changed the initial supply to a `give` instead of  `drip` if the test covers more split specific setups.

### DIFF
--- a/src/test/DripsHub.t.sol
+++ b/src/test/DripsHub.t.sol
@@ -116,22 +116,6 @@ contract DripsHubTest is DripsHubUserUtils {
         collectAll(receiver2, 10);
     }
 
-    function testCollectAllSplitsFundsFromSplits() public {
-        uint32 totalWeight = dripsHub.totalSplitsWeight();
-        setDrips(user, 0, 10, dripsReceivers(receiver1, 10));
-        setSplits(receiver1, splitsReceivers(receiver2, totalWeight));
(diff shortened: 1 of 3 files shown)


src/test/DripsHub.t.sol
```diff
@@ -164,22 +116,6 @@ contract DripsHubTest is DripsHubUserUtils {
         collectAll(receiver2, 10);
     }
 
-    function testCollectAllSplitsFundsFromSplits() public {
```
CodeSandwich, line 167: The `collectAll` is a DripsHub-specific function, is it the right choice to move these tests to `Splits.t.sol`? They don't seem to test there anything more than simple splitting.

src/test/Splits.t.sol
```diff
@@ -0,0 +1,357 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.13;
+
+import {DSTest} from "ds-test/test.sol";
+import {Hevm} from "./Hevm.t.sol";
```
CodeSandwich, line 5: This import can be dropped

CodeSandwich (changes requested):
Really nice work!

CodeSandwich (approved):

PR #138, opened 2022-05-24
Remove drips artificial max timestamp

CodeSandwich:
In the current code the max timestamp was lower than `type(uint64).max` only to cover an absurd edge case of cycle length of 1 second.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -95,6 +95,7 @@ contract DripsHub is Managed {
     /// High value makes collecting cheaper by making it process less cycles for a given time range.
     /// @param _reserve The address of the ERC-20 reserve which the drips hub will work with
     constructor(uint64 _cycleSecs, IReserve _reserve) {
+        require(_cycleSecs > 1, "Cycle length too low");
         cycleSecs = _cycleSecs;
         maxDripsReceivers = Drips.MAX_DRIPS_RECEIVERS;
         maxSplitsReceivers = Splits.MAX_SPLITS_RECEIVERS;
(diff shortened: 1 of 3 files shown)


src/DripsHub.sol
```diff
@@ -95,6 +95,7 @@ contract DripsHub is Managed {
     /// High value makes collecting cheaper by making it process less cycles for a given time range.
     /// @param _reserve The address of the ERC-20 reserve which the drips hub will work with
     constructor(uint64 _cycleSecs, IReserve _reserve) {
+        require(_cycleSecs > 1, "Cycle length too low");
```
xmxanuel, line 98: Where would it break if the cycle secs are one second?

CodeSandwich: Here: https://github.com/radicle-dev/radicle-drips-hub/blob/igor/max_timestamp/src/Drips.sol#L613 In case of a drip which ends after the maximum possible timestamp it'd overflow and revert.

CodeSandwich: An alternative would be to change the notion of a cycle and say that cycle `1` covers funds received not during timestamps `0 - cycleSecs - 1`, but  `cycleSecs - cycleSecs * 2 - 1`. That would have no real-world impact, but would make tests trickier to run, drips would be unusable with `block.timestamp` below `cycleSecs`.

xmxanuel: If `(start + duration )> uint max`

xmxanuel: Ah got it.

xmxanuel (approved):

Issue #128, opened 2022-04-13
Store the last drips balance and update timestamp on-chain


CodeSandwich:
Currently the last drips balance and the last drips update timestamp are passed around as calldata. The user of the contract is expected to retrieve it from the last event and pass it in the next `setDrips` call. I think that we should consider keeping these parameters in storage on-chain.

This would change the drips storage from this:
```solidity
struct DripsState {
    // Slot 1
    bytes32 dripsHash;
    // Slot 2
    uint64 nextCollectedCycle;
    // Slot 3
    mapping(uint64 => AmtDelta) amtDeltas;
}
```

to this:
```solidity
struct DripsState {
    // Slot 1
    bytes32 dripsHash;
    // Slot 2
    uint64 nextCollectedCycle;
    uint128 balance;
    uint64 updated;
    // Slot 3
    mapping(uint64 => AmtDelta) amtDeltas;
}
```

Pros:
- **`HIGH LEVEL`** Easier to use API. These parameters have always been causing confusion. 
- **`HIGH LEVEL`** We gain the possibility of allowing anybody to top-up anybody else's drips balance. We may not choose to do that in DripsHub, but other contracts based on Drips lib could allow that. In the current implementation this is impossible because any top-up changes the drips state hash, so every call to `setDrips` could be front-run and rendered invalid by a top-up call.
- Less transaction calldata and less data passed between contracts. This saves some gas, in top-level TX calldata around 400, I'm not sure how much in in internal TXs.
- **`HIGH LEVEL`** In case there's no `dripsHash` change, there's no need to emit the full drips configuration in the events, the old ones can be easily reused. This is important because events are quite expensive. This could be done in the current implementation, but would require decoupling the drips state hash and the drips configuration hash, that would be complicated to use.
- If https://github.com/radicle-dev/radicle-drips-hub/pull/122 is merged, we may cache on-chain `defaultEnd` without any additional cost, because the slot with `balance` and `updated` needs to be updated anyway.

Whatevers:
- More stable state hashes. Some transactions like drips top-up don't cause the hash state change anymore. In such cases we don't need to update the on-chain `dripsHash`, only the `balance` and `updated`, so still 1 slot.
- The cost of initializing the user's slot with `nextCollectedCycle`, `balance` and `updated` is shared between the user themselves and the first user to drip to them. Globally it doesn't increase the cost.

Cons:
[...]

PR #139, opened 2022-05-24
Store the default drips end on-chain

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-drips-hub/issues/128.

This PR reduces the maximum timestamp from uint64 to uint32 moving the death of drips protocol from year 584554532842 to mere 2106.  It allows storing on-chain the default drips end without using any additional slots.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -73,6 +73,8 @@ library Drips {
         uint32 nextCollectedCycle;
         /// @notice The time when drips have been configured for the last time
         uint32 updateTime;
+        /// @notice The end time of drips without duration
+        uint32 defaultEnd;
         /// @notice The balance when drips have been configured for the last time
         uint128 balance;
         /// @notice The changes of collected amounts on specific cycle.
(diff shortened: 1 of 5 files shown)


src/Drips.sol
```diff
@@ -385,17 +387,17 @@ library Drips {
...
         balance = lastBalance;
         for (uint256 i = 0; i < receivers.length; i++) {
             DripsReceiver memory receiver = receivers[i];
-            uint64 start = receiver.start;
+            uint32 start = receiver.start;
             if (start == 0) start = lastUpdate;
-            uint64 end = defaultEnd;
-            if (receiver.duration != 0) end = _capTimestamp(uint72(start) + receiver.duration);
+            uint32 end = defaultEnd;
+            if (receiver.duration != 0) end = _capTimestamp(uint40(start) + receiver.duration);
```
xmxanuel, line 400: We basically adding to two uint32, so we need more space and `uint40` is the next in line.

This casting with different uint sizes, so many things to consider.  Need to check if there are any nice libraries doing these things correctly.

xmxanuel (approved):
nice good work

PR #140, opened 2022-05-24
Unify drips range calculation logic

CodeSandwich:


xmxanuel (approved):
looks good to me.
The logic seems to work. 

I think we can simplify it a bit more. I opened a PR #151 
to discuss it.

PR #141, opened 2022-05-24
Fix overflow in total amtPerSec

CodeSandwich:


src/Drips.sol
```diff
@@ -357,23 +357,26 @@ library Drips {
...
-        for (uint256 i = 0; i < length; i++) {
-            uint32 start = defaults[i].start;
+        uint136 amtPerSec = 0;
+        uint136 end = type(uint136).max;
+        for (uint8 i = 0; i < length; i++) {
+            DefaultEnd memory defaultEnd = defaults[i];
+            uint32 start = defaultEnd.start;
             if (start >= end) break;
-            balance -= amtPerSec * (start - lastStart);
+            balance -= uint128(amtPerSec * (start - lastStart));
```
xmxanuel, line 373: Do you think it makes sense to helper functions here?
```
safeMul(uint128, uint32)
```
Inside we would cast it to a `uint136` make the multiplication and cast it back?

src/Drips.sol
```diff
@@ -357,23 +357,26 @@ library Drips {
...
             lastStart = start;
-            amtPerSec += defaults[i].amtPerSec;
-            end = _capTimestamp(start + (balance / amtPerSec));
+            amtPerSec += defaultEnd.amtPerSec;
+            end = start + (balance / amtPerSec);
         }
+        if (end > type(uint32).max) end = type(uint32).max;
```
xmxanuel, line 378: Same here

xmxanuel (approved):
looks good to me.

I added some ideas.

src/Drips.sol
CodeSandwich: The point of this PR is to make this exact `amtPerSec` variable a `uint136`, because it needs to be able to handle a sum of all receivers' `amtPerSec`s. We'd need `safeMul(uint136, uint32) returns (uint128)`, which is quite confusing without the context and not very useful. We know that this casting is safe only because if `amtPerSec > balance`, `start == lastStart`, so the result is capped at `balance`, which is `uint128`.

src/Drips.sol
CodeSandwich: The helper casting to `uint32` by capping would be used only once, here, and it'd probably make the logic **harder** to follow.

PR #142, opened 2022-05-25
feat: simplified _updateReceiverStates in Drips

xmxanuel:
Currently, I find it very hard to read the `_updateReceiverStates` method. 

A reader needs to keep a lot of things in his/her mind. Our algorithm is also complex but I gave it a try.

@CodeSandwich 
I tried to simplify it a bit, let me know what you think about it.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -73,6 +73,8 @@ library Drips {
         uint32 nextCollectedCycle;
         /// @notice The time when drips have been configured for the last time
         uint32 updateTime;
+        /// @notice The end time of drips without duration
+        uint32 defaultEnd;
         /// @notice The balance when drips have been configured for the last time
         uint128 balance;
         /// @notice The changes of collected amounts on specific cycle.
@@ -238,8 +240,8 @@ library Drips {
         bytes32 currDripsHash = hashDrips(currReceivers);
         require(currDripsHash == state.dripsHash, "Invalid current drips list");
         uint32 lastUpdate = state.updateTime;
+        uint32 currDefaultEnd = state.defaultEnd;
         uint128 lastBalance = state.balance;
-        uint32 currDefaultEnd = _defaultEnd(lastBalance, lastUpdate, currReceivers);
         {
             uint128 currBalance = _currBalance(
                 lastBalance,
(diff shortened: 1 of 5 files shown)


src/Drips.sol
```diff
@@ -471,24 +473,23 @@ library Drips {
...
                     currStart,
-                    currEnd,
                     newStart,
-                    newEnd,
-                    currRecv.amtPerSec
+                    currEnd,
+                    newEnd
                 );
+                currIdx++;
+                newIdx++;
```
xmxanuel, line 486: I think it is easier to read if the increase of both indexes happens here instead of at the bottom of the while loop.

src/Drips.sol
```diff
@@ -439,12 +439,14 @@ library Drips {
         uint32 cycleSecs = cycleSecs_;
         uint256 currIdx = 0;
         uint256 newIdx = 0;
-        while (true) {
-            bool pickCurr = currIdx < currReceivers.length;
+        bool pickCurr;
+        bool pickNew;
+        do {
```
CodeSandwich, line 444:
```suggestion
        bool pickCurr = true;
        bool pickNew = true;
        while(pickCurr || pickNew) {
```
This makes the context of the loop a little less stretched.

Personally I don't have a problem with `while(true)` loops, they aren't an antipattern, are they? IMO `pickCurr` and `pickNew` are meaningless and confusing outside of the context of an iteration which they describe, it suggests that their values can live across iterations.

CodeSandwich (changes requested):
That's awesome that you're taking a shot at making this logic more accessible for readers!

src/Drips.sol
xmxanuel: I agree that it is definitely not an anti-pattern. I think if the while condition is very complex it can be a lot of times more elegant to use a while(true) together with a break condition.

However, I think here the while condition is not that complex. 

In the end it is taste, here

src/Drips.sol
```diff
@@ -497,17 +498,75 @@ library Drips {
...
+    function _removeDrip(
+        mapping(uint32 => AmtDelta) storage deltas,
+        uint32 cycleSecs,
+        uint128 amtPerSec,
+        uint32 currEnd
+    ) internal {
+        // remove currEnd by adding as a positive delta
+        _setDelta(deltas, cycleSecs, currEnd, int128(amtPerSec));
+        // stop drip by adding negative delta at current timestamp
+        _setDelta(deltas, cycleSecs, _currTimestamp(), -int128(amtPerSec));
```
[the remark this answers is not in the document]
xmxanuel: Ah, you are right. I forgot that case. :+1: 
We need better test coverage for all cases.  A test needs to break in that case.

CodeSandwich:
I think that this PR is now obsolete

xmxanuel:
Yes, I think we might pick up certain optimizations at a later point again.
When we have a more or less feature-complete stable code base. :+1:

PR #143, opened 2022-05-26
Prevent collecting cycles from before the first drip

CodeSandwich:


src/Drips.sol
```diff
@@ -107,10 +107,10 @@ library Drips {
         uint256 userId,
         uint256 assetId
     ) internal view returns (uint32 cycles) {
-        uint32 collectedCycle = s.dripsStates[assetId][userId].nextCollectedCycle;
-        if (collectedCycle == 0) return 0;
+        uint32 nextCollectedCycle = s.dripsStates[assetId][userId].nextCollectedCycle;
         uint32 currFinishedCycle = _currTimestamp() / cycleSecs;
-        return currFinishedCycle + 1 - collectedCycle;
+        if (nextCollectedCycle == 0 || nextCollectedCycle > currFinishedCycle) return 0;
+        return currFinishedCycle - nextCollectedCycle + 1;
```
xmxanuel, line 113: Okay, this makes sense.

Basically, before we had the situation. That `nextCollectedCycle`  could be in future and this method would have resulted in an underflow by substracting `currFinishedCycle` with `collectedCycle`?

src/Drips.sol
```diff
@@ -501,9 +501,10 @@ library Drips {
                     newDefaultEnd
                 );
                 _setDeltaRange(state.amtDeltas, cycleSecs, start, end, newRecv.amtPerSec);
-                // The receiver may have never been used, initialize it
-                if (state.nextCollectedCycle == 0) {
-                    state.nextCollectedCycle = _currTimestamp() / cycleSecs + 1;
+                // Ensure that the receiver collects the updated cycles
+                uint32 startCycle = start / cycleSecs + 1;
```
xmxanuel, line 505: I am wondering if we should back this `+1` operation in some helper method?

src/Drips.sol
```diff
@@ -501,9 +501,10 @@ library Drips {
...
+                if (state.nextCollectedCycle == 0 || state.nextCollectedCycle > startCycle) {
+                    state.nextCollectedCycle = startCycle;
```
xmxanuel, line 507: Uh. That was actually a bug. These scheduled drips.
We should write a test for it.

Something like. 
First add:
Drip 1: starts at timestamp 10

Second add:
Drip 2: starts at timestamp 5

xmxanuel (approved):
Looks good.

We should add a test. Maybe in this PR or adding an issue.

src/Drips.sol
CodeSandwich: No, that was safe, `collectedCycle` was being set to the value equal to `_currTimestamp() / cycleSecs + 1` when initializing it in `setDrips`, so this subtraction could never underflow. That was inefficient but sound.

src/Drips.sol
CodeSandwich: It wasn't a bug, `state.nextCollectedCycle > startCycle` would never happen, because `nextCollectedCycle` was initialized to `_currTimestamp() / cycleSecs + 1`. `startCycle` is capped to never go below `_currTimestamp() / cycleSecs + 1`.

xmxanuel: Ah okay. I somehow thought you changed that.

Would it make sense to change it? Basically, If I want to schedule a drip 10 months into the future. If it is the only drip of the receiver, we wouldn't need to iterate the first 10 months.

src/Drips.sol
CodeSandwich: Fixed in the other PR

src/Drips.sol
CodeSandwich: I'm a little confused about your comment, haven't you just described this exact PR?

xmxanuel: All good. I got confused I somehow thought we set the `nextCollectedCycle` somewhere else and in my mind I only considered the update cases here.

xmxanuel: We forgot the update in the `pickCurr && pickNew` case. We should open a separate PR to fix it.

PR #150, opened 2022-05-30
Clean up the renaming of collect to receive in drips

CodeSandwich:


src/Drips.sol
```diff
@@ -69,18 +69,18 @@ library Drips {
     struct DripsState {
         /// @notice Drips receivers list hash, see `hashDrips`.
         bytes32 dripsHash;
-        /// @notice The next cycle to be collected
-        uint32 nextCollectedCycle;
+        /// @notice The next cycle to be received
+        uint32 nextReceivedCycle;
```
xmxanuel, line 73: Maybe
```suggestion
        uint32 nextReceivableCycle;
```

xmxanuel: Because the receive didn't happen yet

xmxanuel (approved):
lgtm

src/Drips.sol
CodeSandwich: Good point, fixed

PR #151, opened 2022-05-30
fix: simplify drips range

xmxanuel:
I think I was able to simplify the drip range functions a bit.

Currently, the tests are passing but it could be the case that we don't cover one case in the tests.

Basically I removed the `updateTime` because it can be handled with the `startCap` in all cases.

Here is the reasoning. The `startCap` was not used in that past function by setting it to zero. 

For the future range the startCap needs to be currentTime in all cases.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -418,8 +418,8 @@ library Drips {
     /// @param currReceivers The list of the drips receivers set in the last drips update
     /// of the user.
     /// If this is the first update, pass an empty array.
-    /// If this is the first update, pass zero.
     /// @param lastUpdate the last time the sender updated the drips.
+    /// If this is the first update, pass zero.
     /// @param currDefaultEnd Time when drips without duration
     /// were supposed to end according to the last drips update.
     /// @param newReceivers  The list of the drips receivers of the user to be set.
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -73,6 +73,8 @@ library Drips {
         uint32 nextCollectedCycle;
         /// @notice The time when drips have been configured for the last time
         uint32 updateTime;
+        /// @notice The end time of drips without duration
+        uint32 defaultEnd;
         /// @notice The balance when drips have been configured for the last time
         uint128 balance;
         /// @notice The changes of collected amounts on specific cycle.
@@ -238,8 +240,8 @@ library Drips {
         bytes32 currDripsHash = hashDrips(currReceivers);
         require(currDripsHash == state.dripsHash, "Invalid current drips list");
         uint32 lastUpdate = state.updateTime;
+        uint32 currDefaultEnd = state.defaultEnd;
         uint128 lastBalance = state.balance;
-        uint32 currDefaultEnd = _defaultEnd(lastBalance, lastUpdate, currReceivers);
         {
             uint128 currBalance = _currBalance(
                 lastBalance,
@@ -252,7 +254,7 @@ library Drips {
             newBalance = uint128(uint136(balance));
             realBalanceDelta = int128(balance - int128(currBalance));
         }
-        uint32 newDefaultEnd = _defaultEnd(newBalance, _currTimestamp(), newReceivers);
+        uint32 newDefaultEnd = _defaultEnd(newBalance, newReceivers);
         _updateReceiverStates(
             s.dripsStates[assetId],
             cycleSecs,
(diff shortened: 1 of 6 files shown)


src/Drips.sol
```diff
@@ -483,20 +472,12 @@ library Drips {
             } else if (pickCurr) {
                 // Remove an existing drip
                 mapping(uint32 => AmtDelta) storage deltas = states[currRecv.userId].amtDeltas;
-                (uint32 start, uint32 end) = _dripsRangeInFuture(
-                    currRecv,
-                    lastUpdate,
-                    currDefaultEnd
-                );
+                (uint32 start, uint32 end) = _dripsRangeInFuture(currRecv, currDefaultEnd);
```
xmxanuel, line 475: Also here the `lastUpdate` is not needed because the `startCap=currTimestamp` will overwrite it.

src/Drips.sol
```diff
@@ -520,40 +501,36 @@ library Drips {
...
     /// @param endCap The timestamp the drips range end should be capped to
     function _dripsRange(
         DripsReceiver memory receiver,
-        uint32 updateTime,
         uint32 defaultEnd,
         uint32 startCap,
         uint32 endCap
     ) private pure returns (uint32 start, uint32 end_) {
-        start = updateTime;
+        start = startCap;
```
xmxanuel, line 530: @CodeSandwich  I think you are right with your example. We need actually `start=lastUpdate` here. 
However we still could remove the `if (start < updateTime) start = updateTime;` part.

I will add the testcase here for this example

CodeSandwich: I don't think we can do that either, it covers the case of `_dripsRangeInPast` which has `startCap == 0`. T `start != 0`, then nothing would prevent `start < updateTime`.

xmxanuel: Why the `startCap` would be the currentTimestamp.

CodeSandwich: Could you elaborate? `startCap` is current timestamp only in `_dripsRangeInFuture`, I've raised an example of `_dripsInPast`, which has `startCap` of `0`.

xmxanuel: in the `_dripsInPast` case it is now `startCap=lastUpdate`.

CodeSandwich: Ah, right, I've missed that. Let's see what the code looks like after fixing the `currRecv.start == 0 && currRecv.duration != 0.` case. You may be right about the removal of the `if (start < updateTime) start = updateTime;` line.

src/test/Drips.t.sol
```diff
@@ -703,6 +703,25 @@ contract DripsTest is DSTest {
         receiveDrips(receiver, cycleSecs * 2);
     }
 
+    function testUpdateDefaultStartDrip() public {
```
xmxanuel, line 706: @CodeSandwich 
This testcase fails we init `start = startCap;` instead of `start = updateTime;`

CodeSandwich (approved):
I think that it's a sound change :+1:

src/Drips.sol
```diff
@@ -414,11 +414,11 @@ library Drips {
     /// @notice Applies the effects of the change of the drips on the receivers' drips states.
     /// @param states The drips states for a single asset, the key is the user ID
     /// @param cycleSecs_ The cycle length in seconds.
+    /// @param lastUpdate the last time the sender updated the drips.
     /// Must be the same in all calls working on a single storage instance. Must be higher than 1.
     /// @param currReceivers The list of the drips receivers set in the last drips update
     /// of the user.
     /// If this is the first update, pass an empty array.
-    /// @param lastUpdate The timestamp of the last drips update of the user.
```
CodeSandwich, line 421: These changes seemt to be a mistake, they mess up the docs block

src/Drips.sol
```diff
@@ -538,8 +537,8 @@ library Drips {
     /// @notice Calculates the time range in which a receiver is to be dripped to.
     /// This range is capped to provide a view on drips through a specific time window.
     /// @param receiver The drips receiver
-    /// @param updateTime The time when drips are configured
     /// @param defaultEnd The end time of drips without duration
+    /// @param updateTime The time when drips are configured
```
CodeSandwich, line 541: These changes are unneeded, they only break the order of the parameters

xmxanuel: Fixed

PR #152, opened 2022-05-30
Fix nextReceivedCycle not being updated when updating drips

CodeSandwich:


src/Drips.sol
```diff
@@ -476,14 +476,19 @@ library Drips {
...
                     newStart,
                     newEnd,
                     currRecv.amtPerSec
                 );
+                // Ensure that the user receives the updated cycles
+                uint32 startCycle = _cycleOf(newStart, cycleSecs);
+                if (state.nextReceivedCycle > startCycle) {
+                    state.nextReceivedCycle = startCycle;
+                }
             } else if (pickCurr) {
```
xmxanuel, line 492: Basically,  if I remove the earliest drip which is stored in `nextReceivedCycle`. There is nothing we can do because we don't track the next one in the row.

PR #153, opened 2022-05-30
test: fuzz drip receivers list

xmxanuel:
- first fuzz test with a random receiver list
   - setDrips
   - receiveDrips
   - all receiver should be able to receive their drips

### DIFF
--- a/src/test/Drips.t.sol
+++ b/src/test/Drips.t.sol
@@ -6,17 +6,17 @@ import {Hevm} from "./Hevm.t.sol";
 import {Drips, DripsReceiver} from "../Drips.sol";
 
 contract PseudoRandomUtils {
-    bytes32 internal constant SALT_NOT_SET = "salt not set";
-    bytes32 salt = SALT_NOT_SET;
+    bytes32 private constant SALT_NOT_SET = "salt not set";
+    bytes32 private salt = SALT_NOT_SET;
 
     // returns a pseudo-random number between 0 and range
-    function random(uint256 range) internal returns (uint256) {
+    function random(uint256 range) public returns (uint256) {
         require(salt != SALT_NOT_SET, "Salt not set");
         salt = keccak256(bytes.concat(salt));
         return uint256(salt) % range;
     }
 
-    function initSalt(bytes32 salt_) internal {
+    function initSalt(bytes32 salt_) public {
         salt = salt_;
     }
 }
--- a/src/test/Drips.t.sol
+++ b/src/test/Drips.t.sol
@@ -135,7 +135,7 @@ contract DripsTest is DSTest {
 
         storeCurrReceivers(assetId, userId, newReceivers);
         assertEq(newBalance, balanceTo, "Invalid drips balance");
-        (, uint32 updateTime, uint128 actualBalance) = Drips.dripsState(s, userId, assetId);
+        (, uint32 updateTime, uint128 actualBalance, ) = Drips.dripsState(s, userId, assetId);
         assertEq(updateTime, block.timestamp, "Invalid new last update time");
         assertEq(balanceTo, actualBalance, "Invalid drips balance");
         assertEq(realBalanceDelta, balanceDelta, "Invalid real balance delta");
@@ -146,7 +146,7 @@ contract DripsTest is DSTest {
         uint256 userId,
         DripsReceiver[] memory currReceivers
     ) internal {
-        (bytes32 actual, , ) = Drips.dripsState(s, userId, assetId);
+        (bytes32 actual, , , ) = Drips.dripsState(s, userId, assetId);
         bytes32 expected = Drips.hashDrips(currReceivers);
         assertEq(actual, expected, "Invalid drips configuration");
     }
--- a/src/test/Drips.t.sol
+++ b/src/test/Drips.t.sol
@@ -106,6 +106,78 @@ contract DripsTest is DSTest {
         return recv(recv(recv1, recv2, recv3), recv4);
     }
 
+    function random(
+        bytes32 name,
+        uint256 salt,
+        uint256 range
+    ) internal pure returns (uint256) {
+        return uint256(keccak256(abi.encode(name, salt))) % range;
+    }
+
+    function combineSalt(uint256 a, uint256 b) public pure returns (uint256) {
+        return uint256(keccak256(abi.encode(a, b)));
+    }
+
+    function genRandomRecv(
+        uint256 randomSalt,
+        uint8 amountReceiver,
+        uint128 maxAmtPerSec,
+        uint32 maxStart,
+        uint32 maxDuration
+    ) internal pure returns (DripsReceiver[] memory) {
+        uint256 inPercent = 100;
+        uint256 probDefaultEnd = random("default.end", randomSalt, inPercent);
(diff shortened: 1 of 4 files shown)


src/test/Drips.t.sol
```diff
@@ -106,6 +106,82 @@ contract DripsTest is DSTest {
...
+        uint32 updateTime
+    ) internal {
+        emit log_named_uint("defaultEnd:", defaultEnd);
+        for (uint256 i = 0; i < receivers.length; i++) {
+            DripsReceiver memory r = receivers[i];
+            uint32 duration = r.duration;
+            uint32 start = r.start;
+            if (start == 0) start = updateTime;
+            if (duration == 0) duration = defaultEnd - start;
+            if (start < updateTime) duration -= updateTime - start;
```
xmxanuel, line 161: I tried to do these checks a bit differently compared with the dripsRange function.

Obviously,  we could have also called dripsRange function here but I think it is good to do it a bit different to catch potential errors in dripsRange function.

xmxanuel: I will run it with multiple fuzz runs tonight. Just in case I forgot one of the cases here.

src/test/Drips.t.sol
```diff
@@ -213,6 +289,15 @@ contract DripsTest is DSTest {
         Drips.setDrips(s, cycleSecs, userId, assetId, currReceivers, balanceDelta, newReceivers);
     }
 
+    function receiveDripsExternal(
```
CodeSandwich, line 292: I don't think that this function is used anywhere

src/test/Drips.t.sol
```diff
@@ -106,6 +106,82 @@ contract DripsTest is DSTest {
...
+        DripsReceiver[] memory receivers;
+        for (uint8 i = 0; i < amountReceiver; i++) {
+            uint256 salt = combineSalt(randomSalt, i);
+            uint256 amtPerSec = random("amtPerSec", salt, maxAmtPerSec) + 1;
+            uint256 start = random("start", salt, maxStart);
+            if (start % 100 <= probStartNow) start = 0;
+            uint256 duration = random("duration", salt, maxDuration);
+            if (duration % 100 <= probDefaultEnd) duration = 0;
+
+            receivers = recv(receivers, recv(uint256(i), amtPerSec, start, duration));
```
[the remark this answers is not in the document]
xmxanuel: Yes, I mean it is clear that it is slower but I thought it would be insignificant for the tests.  However, you are right with a lot of receivers and a high number of fuzz runs. 

I guess I just wanted to use your nice helper functions.  :smiley_cat: 

I will change it.

src/test/Drips.t.sol
```diff
@@ -106,6 +106,82 @@ contract DripsTest is DSTest {
         return recv(recv(recv1, recv2, recv3), recv4);
     }
 
+    function random(bytes32 name, uint256 salt) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
xmxanuel: fixed

src/test/Drips.t.sol
```diff
@@ -106,6 +106,78 @@ contract DripsTest is DSTest {
...
     }
 
+    function random(
+        bytes32 name,
+        uint256 salt,
+        uint256 range
+    ) internal pure returns (uint256) {
+        return uint256(keccak256(abi.encode(name, salt))) % range;
+    }
+
```
CodeSandwich, line 116: Maybe something like that would suffice?

```solidity
bytes32 constant internal SEED_NOT_SET = "seed not set";
bytes32 seed = SEED_NOT_SET;

function random(uint256 range) internal returns (uint256) {
    require(seed != SEED_NOT_SET, "Seed not set"); 
    seed =  keccak256(bytes.concat(seed));
    return seed % range;
}

function testFuzzDripsReceiver(bytes32 fuzzSeed) public {
    seed = fuzzSeed;
    ...
}
```
The test run would be still easily reproducible, just feed the test with the same seed, but we would get rid of the awkward random numbers generation. It'd be also impossible to use `random` wrong, it'd always generate a fresh random number.

CodeSandwich (approved):
Nice! I basically have a single suggestion around randomness.

src/test/Drips.t.sol
```diff
@@ -106,6 +106,78 @@ contract DripsTest is DSTest {
...
+        uint128 maxAmtPerSec,
+        uint32 maxStart,
+        uint32 maxDuration
+    ) internal pure returns (DripsReceiver[] memory) {
+        uint256 inPercent = 100;
+        uint256 probDefaultEnd = random("default.end", randomSalt, inPercent);
+        uint256 probStartNow = random("start.now", randomSalt, inPercent);
+        DripsReceiver[] memory receivers = new DripsReceiver[](amountReceiver);
+        for (uint8 i = 0; i < amountReceiver; i++) {
+            uint256 salt = combineSalt(randomSalt, i);
```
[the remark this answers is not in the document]
xmxanuel: Yes, first I wanted to do the same thing but the fuzzing will test `type(uint).max` for sure. 

So it was easier like that. We could also do it as you proposed with an `unchecked`

src/test/Drips.t.sol
```diff
@@ -106,6 +106,78 @@ contract DripsTest is DSTest {
         return recv(recv(recv1, recv2, recv3), recv4);
     }
 
+    function random(
+        bytes32 name,
+        uint256 salt,
```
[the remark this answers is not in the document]
xmxanuel: I looked up the definition from wiki, I got curious. 

> In [cryptography](https://en.wikipedia.org/wiki/Cryptography), a salt is [random](https://en.wikipedia.org/wiki/Random_Number_Generator) data that is used as an additional input to a [one-way function](https://en.wikipedia.org/wiki/One-way_function) that [hashes](https://en.wikipedia.org/wiki/Cryptographic_hash_function) [data](https://en.wikipedia.org/wiki/Data_(computing)), a [password](https://en.wikipedia.org/wiki/Password) or [passphrase](https://en.wikipedia.org/wiki/Passphrase).
Mhm, for me it is fine. We use it a bit differently but in general, it is additional input for the hash function.

src/test/Drips.t.sol
xmxanuel: Nice idea. I will add it. :+1: 

It will be a bit easier to use

src/test/Drips.t.sol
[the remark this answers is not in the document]
xmxanuel: Not needed anymore with the new implementation of `random`

src/test/Drips.t.sol
```diff
@@ -5,7 +5,23 @@ import {DSTest} from "ds-test/test.sol";
 import {Hevm} from "./Hevm.t.sol";
 import {Drips, DripsReceiver} from "../Drips.sol";
 
-contract DripsTest is DSTest {
+contract PseudoRandomUtils {
+    bytes32 private constant SALT_NOT_SET = "salt not set";
+    bytes32 private salt = SALT_NOT_SET;
```
CodeSandwich, line 10: Wise move with encapsulation of the randomness logic! In that case we don't even need the `SALT_NOT_SET` hack in hopes that the salt will never be genuinely set to `salt not set`, we can have a boolean flag for that.

CodeSandwich (approved):

PR #154, opened 2022-06-03
Sort default ends with heap sort

CodeSandwich:
Replaces insert sort with heap sort when calculating the default drips end time.

For the extreme case of 100 random items the cost of sorting itself is cut from ~1,680,000 gas to ~229,000 gas.

I wonder if we need some benchmarking tests to prevent future regressions and how should we organize them?

The representation of a default end is changed from a structure with 2 fields to a packed uint256 to reduce gas. Unfortunately a user defined type did introduce some overhead despite being a theoretically zero-cost abstraction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,12 @@
     "license": "GPL-3.0-only",
     "devDependencies": {
         "prettier": "^2.4.0",
-        "prettier-plugin-solidity": "^1.0.0-beta.18",
-        "solhint": "^3.3.6",
+        "prettier-plugin-solidity": "https://github.com/prettier-solidity/prettier-plugin-solidity.git#fc77fd6eff0c232d84ff50a147592a94b62e03d7",
+        "solhint": "^3.3.7",
         "solhint-plugin-prettier": "^0.0.5"
+    },
+    "resolutions": {
+        "solhint/@solidity-parser/parser": "0.14.2-beta.1"
     },
     "scripts": {
         "lint": "solhint --max-warnings=0 'src/**/*.sol'",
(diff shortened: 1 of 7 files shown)


xmxanuel:
Is your benchmark code available or do we already have a 100 random receiver test case?

Did you compare it with an insert sort algorithm where you first generate the defaultEnds array and sort afterward with insertSort?

I don't know if the additional code complexity is worth the gas optimization. 

Mainly because we can do it off-chain, which will also result in a gas improvement. 

I am starting with the high level before I dive into reviewing it line by line.

CodeSandwich:
Maybe a good move would be to extract the default end calculation logic into a separate library. From the outside it'd be an invisible change, but it'd let us write some tests and benchmarks directly hammering the algorithm :thinking:

CodeSandwich:
That could be because `maxStart` is just 100 while the current timestamp is 10 meaning that there are a lot of overlapping entries. I'm not sure what `probDefaultEnd` does, but it may also reduce the number of receivers with default ends. Also half a million gas is already a pretty good result anyway.

CodeSandwich:
Extracted a default ends heap library, added some tests for it.

The reported gas usage is higher than what was in `igor/sort`. This is partially because the gas is measured more thoroughly, but even when accounting for that the usage has risen by ~10%. This is the effect of addition of `peekStart` allowing skipping an unnecessary `popAmtPerSec` in real world scenarios when not all elements are consumed.

CodeSandwich:
Let's pause it for now as https://github.com/radicle-dev/drips-contracts/pull/160 will probably make this PR obsolete.

PR #157, opened 2022-06-08
Pack the drips parameters

### LINKED ISSUE #145: Pack the drips parameters
# The current situation

Currently the `DripsReceiver` parameter is defined as:

```solidity
struct DripsReceiver {
    uint256 userId;
    uint128 amtPerSec;
    uint32 start;
    uint32 duration;
}
```

And `DripsReceiverSeen` event as:

```solidity
event DripsReceiverSeen(
    bytes32 indexed receiversHash,
    uint256 indexed userId,
    uint128 amtPerSec,
    uint32 start,
    uint32 duration
);
```

# The proposal

Pack all 3 of `amtPerSec`, `start` and `duration` into a single `uint256` word as

```
uint256 dripParams = amtPerSec << 64 | start << 32 | duration;
```

Technically we use a [User Defined Value Type](https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types) for this.

## The advantages

- Lower events gas cost. Each emitted `DripsReceiverSeen` has gas usage reduced by 512, from 1893 to 1381

- Lower calldata gas cost. Each `DripsReceiver` in calldata has its maximum gas usage reduced by 256, from 1184 to 928. Additionally there's less to pass around in memory and hash. Bit-shifting is very cheap, so accessing fields is marginally more expensive.

- Easier sorting of the receivers list. Instead of having a 4-step comparison (`userId` => `amtPerSec` => `start` => `duration`) we have a 2-step (`userId` => `dripParams`).

## The disadvantages

- Harder to work with API, requires packing and unpacking parameters.

CodeSandwich:
Closes #145.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -5,14 +5,60 @@ pragma solidity ^0.8.13;
 struct DripsReceiver {
     /// @notice The user ID.
     uint256 userId;
-    /// @notice The amount per second being dripped. Must never be zero.
-    uint128 amtPerSec;
-    /// @notice The timestamp when dripping should start.
+    /// @notice The drips configuration.
+    DripsConfig config;
+}
+
+/// @notice Describes a drips configuration.
+/// It's constructed from `amtPerSec`, `start` and `duration` as
+/// `amtPerSec << 64 | start << 32 | duration`.
+/// `amtPerSec` is the amount per second being dripped. Must never be zero.
+/// `start` is the timestamp when dripping should start.
+/// If zero, use the timestamp when drips are configured.
+/// `duration` is the duration of dripping.
+/// If zero, drip until balance runs out.
+type DripsConfig is uint192;
+
+using DripsConfigImpl for DripsConfig global;
+
+library DripsConfigImpl {
+    /// @notice Create a new DripsConfig.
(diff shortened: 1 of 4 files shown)


src/Drips.sol
```diff
@@ -5,14 +5,60 @@ pragma solidity ^0.8.13;
...
+    }
+
+    /// @notice Extracts duration from a `DripsConfig`
+    function duration(DripsConfig config) internal pure returns (uint32) {
+        return uint32(DripsConfig.unwrap(config));
+    }
+
+    /// @notice Compares two `DripsConfig`s.
+    /// First compares their `amtPerSec`s, then their `start`s and then their `duration`s.
+    function lt(DripsConfig config, DripsConfig otherConfig) internal pure returns (bool) {
```
xmxanuel, line 59: No operator overloading in Solidity :)

xmxanuel (approved):
Nice work :+1: 

I would add some unit tests for the library.

This might also help readers to understand what's going on, if you are not so familiar with bit manipulation, shifting and casting.

CodeSandwich:
Added tests and rebased to `manuel/defaultEndFormular`

PR #163, opened 2022-06-21
Add balanceAt

CodeSandwich:


src/Drips.sol
```diff
@@ -213,6 +213,29 @@ library Drips {
...
+    /// @param s The drips storage
+    /// @param userId The user ID
+    /// @param assetId The used asset ID
+    /// @param receivers The current drips receivers list
+    /// @param timestamp The timestamps for which balance should be calculated.
+    /// It can't be lower than the timestamp of the last call to `setDrips`.
+    /// If it's bigger than `block.timestamp`, then it's a prediction assuming
+    /// that `setDrips` won't be called before `timestamp`.
+    /// @return balance The user balance on `timestamp`
+    function balanceAt(
```
xmxanuel, line 226: Should we add a second function `balanceAt` without the `timestamp` parameter and we call the other `balanceAt`  with `timestamp=block.timestamp` ?

src/test/Drips.t.sol
```diff
@@ -228,7 +229,43 @@ contract DripsTest is DSTest, PseudoRandomUtils {
     }
 
     function assertBalance(uint256 userId, uint128 expected) internal {
-        changeBalance(userId, expected, expected);
+        assertBalanceAt(userId, expected, block.timestamp);
+    }
+
+    function assertBalanceAt(
```
xmxanuel, line 235: :clap:

xmxanuel (approved):
nice. I think that's a good idea.

src/Drips.sol
CodeSandwich: Hmm, I think that it wouldn't be worth the public API surface increase. For on-chain callers explicitly adding a `block.timestamp` parameter isn't a big inconvenience. For off-chain callers it'd be useless, because they wouldn't know the timestamp for which the balance is valid. They'd need to fetch the current block timestamp and then it's safer to just use that timestamp as a parameter.

xmxanuel: If you call a view function off-chain, it will take the latest block timestamp  for the block.timestamp variable. 

My reasoning was that this easier way and less error-prone. Then let a user figure out the correct format of the timestamp.

xmxanuel: I still remember that you acutally believed a block.timestamp used the seconds passed from the first minted block on Ethereum instead of the UNIX timestamp. 

However, no strong opinion here.

CodeSandwich: > If you call a view function off-chain, it will take the latest block timestamp for the block.timestamp variable.

Yes, but let's say that you got a result `50`, but what timestamp does it describe? It certainly doesn't say anything about the present or the future, it's a slightly obsolete state. I'm fine with adding a function to the SDK which queries for the current UNIX timestamp, but I think that it's not worth adding to the public smart contract API, it should be as lean as possible.

xmxanuel: Yeah, that is also true. I think adding it to the SDK is a good idea.
Then let's not add it. :+1:

CodeSandwich: > I still remember that you acutally believed a block.timestamp used the seconds passed from the first minted block on Ethereum instead of the UNIX timestamp.

:sweat_smile: But seriously, if a user of the API doesn't know that, they won't be able to even `setDrips`, because it requires that knowledge. We can't shield users from misusing the API by bloating it and creating permutations of defaults, quite the contrary, we must keep it simple, there are already a lot of functions.

PR #165, opened 2022-06-24
Rename account to app in DripsHub

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/162. CC: @gh0stwheel

The PR also contains some final cleanups of the v0.1 sub-accounts, it was finally easy to catch the last few mentions of them and even an obsolete test.

### DIFF
--- a/src/AddressApp.sol
+++ b/src/AddressApp.sol
@@ -4,7 +4,14 @@ pragma solidity ^0.8.13;
 import {DripsHub, DripsReceiver, IERC20, SplitsReceiver} from "./DripsHub.sol";
 import {SafeERC20} from "openzeppelin-contracts/token/ERC20/utils/SafeERC20.sol";
 
-contract AddressId {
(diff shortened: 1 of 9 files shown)


xmxanuel:
I think it would be nice to also rename the `AddressId` to `AddressDrips`.

CodeSandwich:
But what about splits? `AddressId` is wide enough to cover the whole protocol, `AddressDrips` suggests that it can be used only for identification of drips.

xmxanuel:
AddressApp is also very good. I just want to increase the probablility if
you want to build on Drips that you open the right contract.

Igor Żuk ***@***.***> schrieb am Mo., 27. Juni 2022, 11:48:

CodeSandwich:
Renamed `AddressId` to `AddressApp`, @xmxanuel

xmxanuel (approved):

PR #166, opened 2022-06-28
Add deploy script

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/161.

A lightly modified script from `radicle-periphery`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,123 @@ dapp test -m <REGEX>
 dapp test -m testName
 dapp test -m ':ContractName\.'
 ```
+
+## Deploy to a local testnet
+Start a local testnet node and let it run in the background:
+
+```bash
+dapp testnet
+```
+
+Set up environment variables.
+See instructions for public network deployment to see all the options.
+To automatically set bare minimum environment variables run:
+
+```bash
+source scripts/local-env.sh
+```
+
+Run deployment:
+
+```bash
+scripts/deploy.sh
+```
+
+## Deploy to a public network
+
+Use dapp.tools' `ethsign` to query or add keys available on the system for signing transactions.
+
+Set up environment variables controlling the deployment process:
+
+```bash
+# The RPC URL to use, e.g. `https://mainnet.infura.io/MY_INFURA_KEY`.
+# Contracts will be deployed to whatever network that endpoint works in.
+export ETH_RPC_URL="<URL>"
+
+# Address to use for deployment. Must be available in `ethsign`
+export ETH_FROM="<ADDRESS>"
+
+# OPTIONAL
+# The file containing password to decrypt `ETH_FROM` private key from keystore.
+# If not set, the password will be prompted multiple times during deployment.
+# If `ETH_FROM` is not password protected, can be set to `/dev/null`.
+export ETH_PASSWORD="<KEYSTORE_PASSWORD>"
+
+# OPTIONAL
+# The API key to use to submit contracts' code to Etherscan.
+# In case of deployments to networks other than Ethereum an appropriate equivalent service is used.
+export ETHERSCAN_API_KEY="<KEY>"
+
+# OPTIONAL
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -32,3 +32,120 @@ dapp test -m <REGEX>
...
+
+```bash
+
+# OPTIONAL
+# Address of Reserve to use. If not set, a new instance is deployed.
+export RESERVE="<ADDRESS>"
+
+# OPTIONAL
+# Address of Reserve owner to set. If not set, `ETH_FROM` is used.
+export RESERVE_OWNER="<ADDRESS>"
```
xmxanuel, line 97: I think technically the `RESERVE_OWNER` and the `DRIPS_HUB_ADMIN` could be different addresses.

However, it might be the same in most cases.

One `GOVERNANCE` might be easier?

scripts/deploy.sh
```diff
@@ -0,0 +1,136 @@
...
+# Build the contracts
+message Building Contracts
+dapp build
+
+# Deploy the contracts
+message Deploying contracts
+
+if [ -z "$RESERVE" ]; then
+    deploy "Reserve" 'src/Reserve.sol:Reserve' "$ETH_FROM"
+    RESERVE=$DEPLOYED_ADDR
```
xmxanuel, line 76: Is it not possible to return a string from a bash function in an elegant way?

scripts/deploy.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo
+fi
+
+if [ $(seth call "$RESERVE" 'owner()(address)') != $(seth --to-address $RESERVE_OWNER) ]; then
+    echo -e "Setting Reserve owner to $RESERVE_OWNER\n"
+    seth send "$RESERVE" 'transferOwnership(address)()' "$RESERVE_OWNER"
+    echo
+fi
+
+if [ $(seth call "$DRIPS_HUB" 'admin()(address)') != $(seth --to-address $DRIPS_HUB_ADMIN) ]; then
```
xmxanuel, line 107: I think this can never be false.

Because in the deployment of DRIPS_HUB you defined the admin already as `DRIPS_HUB_ADMIN`

xmxanuel (changes requested):
Overall looks good. I found some minor errors in the script.

scripts/deploy.sh
CodeSandwich: The short answer is no, bash is a really terrible **programming** language :laughing: One often workaround is to `echo` the returned value in the function and then `$()` it on the caller side, this is what `drips-periphery` version does. Unfortunately it makes printing to the console from the inside of the function a little tricky, it can't use stdout anymore and it and all it can do is (ab)use stderr. I find a magic variable a little easier to control. It's also not unheard of in the bash world, many function do that, e.g. the built-in `read`.

scripts/deploy.sh
CodeSandwich: It can, if the user provides `DRIPS_HUB` themselves, it bypasses the deployment and setting to `DRIPS_HUB_ADMIN`. It's an edge case, but I think that it makes the script a little easier to handle by guaranteeing to always deliver what the user asks for.

README.md
CodeSandwich: That's a good point, added `GOVERNANCE`.

scripts/deploy.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo -e "Setting DripsHub admin to $DRIPS_HUB_ADMIN\n"
+    seth send "$RESERVE" 'transferOwnership(address)()' "$DRIPS_HUB_ADMIN"
```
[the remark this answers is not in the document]
CodeSandwich: Good catch! Fixed.

scripts/deploy.sh
xmxanuel: Ah, I see. Thanks for sharing.
I think it is okay. The script is not so complex and one can spot the pattern.

xmxanuel (approved):
looks good

PR #169, opened 2022-07-18
Switch default end calculation to binary search 

CodeSandwich:
Part of https://github.com/radicle-dev/drips-contracts/issues/167 and https://github.com/radicle-dev/drips-contracts/issues/159.

This PR requires an update to prettier, because the old version kept removing the `memory-safe` annotation.


xmxanuel (approved):
LGTM.

PR #170, opened 2022-07-18
Merge dev/2.0

CodeSandwich:
It's about time to finally merge `dev/2.0` into `master`, because we're nearing the official release of v0.2. The current state of v0.1 is safe in branch `master_v0.1`.

CC: @gh0stwheel @evvvritt


gh0stwheel:
Awesome! 💦 💦 💦 💦

xmxanuel (approved):
Nice :100:

I think we should start making `releases` then there is no need for a `master_v1`.

gh0stwheel:
> I think we should start making releases then there is no need for a master_v1.

I had a similar thought. This might be a good time to take a look at our branch/merge process as a team and think about what we want to use for the smart contracts going forwad.

I also had the thought that it might make sense to use tags for when we make network deploys so that those versions are always clear and easy to find. WDYT?

CodeSandwich:
I've proposed using a **branch** and not just a tag because the contracts are upgradable and theoretically we may still want to develop them and e.g. publish bugfixes. I agree, marking releases with tags is a great idea!

PR #177, opened 2022-07-27
Migrate to Foundry

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/81


xmxanuel (approved):
Nice. Looks good.
I think it is the right time to migrate to foundry.

scripts/deploy.sh
```diff
@@ -1,50 +1,52 @@
 #! /usr/bin/env bash
 
-set -e
-
-message() {
+set -eo pipefail
 
+print_title() {
```
xmxanuel, line 5: :kissing:

PR #200, opened 2022-08-25
Add eip-2612 support

ghost:
Closes #192

PR #201, opened 2022-08-25
Refactor style and update docs

ghost:
Depends on #200


CodeSandwich:
I think that most of the changes introduced in this PR should be a part of #200, WDYT?

PR #203, opened 2022-09-02
Add caller

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/192.
Makes https://github.com/radicle-dev/drips-contracts/issues/183 more reasonable.
Makes https://github.com/radicle-dev/drips-contracts/issues/103 more feasible if we choose to go with the `AutomationApp` approach.

It's a minimalistic solution, which seems to still manage to cover all the use cases which we need while staying simple and easily auditable.

It follows ERC-2771, so on the app side all that's needed is deriving from [ERC2771Context](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol) and then using `_msgSender()` instead of `msg.sender`. See changes made to `AddressApp`.

### An example of how a signature for `callSigned` can be generated with ethers.js:
```ts
// https://codesandbox.io/s/typescript-playground-export-forked-feumq1

import * as ethers from "ethers";

async function main() {
  const wallet = new ethers.Wallet(ethers.utils.hexZeroPad("0x01", 32));
  const sender = wallet.address;
  console.log("Address: " + sender);

  const domain = {
    name: "Caller",
    version: "1",
    chainId: 31337,
    verifyingContract: "0xCe71065D4017F316EC606Fe4422e11eB2c47c246"
  };
  const types = {
    CallSigned: [
      { type: "address", name: "sender" },
      { type: "address", name: "to" },
      { type: "bytes", name: "data" },
      { type: "uint256", name: "value" },
      { type: "uint256", name: "nonce" },
      { type: "uint256", name: "deadline" }
    ]
  };
  // If possible, use the interfaces generated from the ABIs
  const iface = new ethers.utils.Interface(["function foo(uint256 input)"]);
  const value = {
    sender: sender,
    to: "0x185a4dc360CE69bDCceE33b3784B0282f7961aea",
    data: iface.encodeFunctionData("foo", [1234567890]),
    value: 4321,
    nonce: 0,
    deadline: ethers.constants.MaxUint256.toString()
  };
  const signatureRaw = await wallet._signTypedData(domain, types, value);
  const signature = ethers.utils.splitSignature(signatureRaw);
  console.log("r: " + signature.r);
  console.log("sv: " + signature.yParityAndS);
}

main();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,11 @@ Set up environment variables configuring the deployed contracts:
 export GOVERNANCE="<ADDRESS>"
 
 # OPTIONAL
+# Address of Caller to use. If not set, a new instance is deployed.
+# If set to zero, newly deployed Caller-aware contracts don't get support for Caller.
+export CALLER="<ADDRESS>"
+
+# OPTIONAL
 # Address of Reserve to use. If not set, a new instance is deployed.
 export RESERVE="<ADDRESS>"
 
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -50,6 +50,7 @@ fi
 echo "Etherscan API key:        $ETHERSCAN_API_KEY_PROVIDED"
 echo "Deployment JSON:          $DEPLOYMENT_JSON"
 TO_DEPLOY="to be deployed"
+echo "Caller:                   ${CALLER:-$TO_DEPLOY}"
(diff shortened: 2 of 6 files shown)


gh0stwheel:
This is super elegant. Nice work @CodeSandwich !

PR #233, opened 2022-11-23
Make metadata key bytes32

CodeSandwich:


src/NFTDriver.sol
```diff
@@ -168,7 +168,7 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Upgradeable {
     /// The token ID is equal to the user ID controlled by it.
     /// @param key The metadata key
     /// @param value The metadata value
-    function emitUserMetadata(uint256 tokenId, uint256 key, bytes calldata value)
+    function emitUserMetadata(uint256 tokenId, bytes32 key, bytes calldata value)
```
xmxanuel, line 171: I am wondering if we should at `emit` a second event on this layer with the current holder of the `tokenId`

Basically
```solidity
function emitUserMetadata(uint256 tokenId, bytes32 key, bytes calldata value) { 
  emit NFTMetaData(ownerOf(tokenId), tokenId);
  dripsHub.emitUserMetadata(tokenId, key, value);
```

It wouldn't be that expensive but could be useful if later Apps build on the Metadata feature.
That it is easy to detected which `owner` emitted new metaData etc.

xmxanuel (approved):
looks good.

One idea for a second event on the Driver layer.

src/NFTDriver.sol
CodeSandwich: ERC-721 already has `event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);` which is emitted when minting, transferring, and finally burning a token, how is `even NFTMetadata(address indexed owner, uint256 indexed tokenId)` better?

PR #234, opened 2022-11-23
Clean up returned values

CodeSandwich:
The changes:
- `receiveDrips`, `receiveDripsResult` - stop returning `receivableCycles`, because it's easily accessible via `receivableDripsCycles`.
- `splitResults` - return `splitAmt` to consistently match `split` API, even though `splitAmt` could be calculated as `amount - collectableAmt`, but that's confusing.
- `setDrips` - stop returning `newBalance`, because it's easily accessible via `dripsState` along the rest of the state.

### DIFF
--- a/src/DripsHub.sol
+++ b/src/DripsHub.sol
@@ -277,11 +277,13 @@ contract DripsHub is Managed, Drips, Splits {
     /// @param userId The user ID
     /// @param currReceivers The list of the user's current splits receivers.
     /// @param amount The amount being split.
-    /// @return collectableAmt The part of the `amount` left for collection after splitting.
+    /// @return collectableAmt The amount made collectable for the user
+    /// on top of what was collectable before.
+    /// @return splitAmt The amount split to the user's splits receivers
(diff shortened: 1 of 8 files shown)


src/DripsHub.sol
```diff
@@ -207,14 +205,13 @@ contract DripsHub is Managed, Drips, Splits {
     /// If too low, receiving will be cheap, but may not cover many cycles.
     /// If too high, receiving may become too expensive to fit in a single transaction.
     /// @return receivedAmt The received amount
-    /// @return receivableCycles The number of cycles which still can be received
     function receiveDrips(uint256 userId, IERC20 erc20, uint32 maxCycles)
         public
         whenNotPaused
-        returns (uint128 receivedAmt, uint32 receivableCycles)
+        returns (uint128 receivedAmt)
```
xmxanuel, line 211: I think it makes sense. As you said the user could use `receivableDripsCycles` to get the cycles. 
It is also kind of an edge-case, that a user needs this feature. 

So it makes sense to remove the information for the `receiveDrips` :+1:

PR #238, opened 2022-11-25
Clean up Drips API

CodeSandwich:
Clean up the internal API. Remove return values not used in DripsHub and make `_calcMaxEnd` private.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -221,11 +221,11 @@ abstract contract Drips {
     /// If too low, receiving will be cheap, but may not cover many cycles.
     /// If too high, receiving may become too expensive to fit in a single transaction.
     /// @return receivedAmt The received amount
-    /// @return receivableCycles The number of cycles which still can be received
     function _receiveDrips(uint256 userId, uint256 assetId, uint32 maxCycles)
         internal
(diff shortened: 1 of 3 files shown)


src/Drips.sol
```diff
@@ -605,7 +605,7 @@ abstract contract Drips {
         DripsReceiver[] memory receivers,
         uint32 maxEndTip1,
         uint32 maxEndTip2
-    ) internal view returns (uint32 maxEnd) {
+    ) private view returns (uint32 maxEnd) {
```
xmxanuel, line 608: Okay make sense because we are using internal types like `DripsReceiver`

src/DripsHub.sol
```diff
@@ -241,7 +241,7 @@ contract DripsHub is Managed, Drips, Splits {
         returns (uint128 receivedAmt)
     {
         uint256 assetId = _assetId(erc20);
-        (receivedAmt,) = Drips._receiveDrips(userId, assetId, maxCycles);
+        receivedAmt = Drips._receiveDrips(userId, assetId, maxCycles);
```
xmxanuel, line 244: nicer

xmxanuel (approved):

PR #239, opened 2022-11-29
Improve maxEnd tests

CodeSandwich:


test/Drips.t.sol
```diff
@@ -263,35 +263,58 @@ contract DripsTest is Test, PseudoRandomUtils, Drips {
...
+
     function setDrips(
         uint256 userId,
         uint128 balanceFrom,
         uint128 balanceTo,
-        DripsReceiver[] memory newReceivers
+        DripsReceiver[] memory newReceivers,
+        uint256 expectedMaxEndFromNow
+    ) internal {
+        setDrips(userId, balanceFrom, balanceTo, newReceivers, 0, 0, expectedMaxEndFromNow);
```
xmxanuel, line 277: Would it not make sense if the `expectedMaxEndFromNow` is known to pass two tips?
`expectedMaxEndFromNow, expectedMaxEndFromNow+1`?
(Assuming here expectedMaxEndFromNow last timestamp with enough balance)

xmxanuel: I mean it is a test so doesn't really matter :) If we don't test the `tips` here in specific

test/Drips.t.sol
```diff
@@ -1403,59 +1421,130 @@ contract DripsTest is Test, PseudoRandomUtils, Drips {
...
+        DripsReceiver[] memory receivers,
+        uint32 maxEndTip1,
+        uint32 maxEndTip2,
+        uint256 expectedMaxEndFromNow
+    ) internal {
+        emit log_named_uint("Setting drips with tip 1", maxEndTip1);
+        emit log_named_uint("               and tip 2", maxEndTip1);
+        uint256 snapshot = vm.snapshot();
+        setDrips(sender, 0, amt, receivers, maxEndTip1, maxEndTip2, expectedMaxEndFromNow);
+        vm.revertTo(snapshot);
```
xmxanuel, line 1542: This is cool. I never used `snapshot` so far. :+1:

test/Drips.t.sol
```diff
@@ -1403,59 +1421,130 @@ contract DripsTest is Test, PseudoRandomUtils, Drips {
...
+        setDripsPermuteTips({
+            amt: 10,
+            receivers: recv(receiver, 1),
+            maxEndTip1: type(uint32).max,
+            maxEndTip2: 25,
+            expectedMaxEndFromNow: 10
+        });
+    }
+
+    function setDripsPermuteTips(
```
xmxanuel, line 1513: nice idea to test all permutations in that way

xmxanuel (approved):
nice work on the tests

test/Drips.t.sol
CodeSandwich: Yeah, the gas savings aren't important, it's much more important to keep the code simple. Without tips the tests are also more thorough, because the max end search must be done over the entire space, so the code coverage should be a bit better, and more edge cases may be caught.

test/Drips.t.sol
CodeSandwich: It's super useful, it was even present in the hardhat EVM, it'd be difficult to build a test framework without an easy way to restore the initial state.

PR #244, opened 2022-12-12
Add dripsHistoryHashes to SqueezedDrips event

CodeSandwich:
Fixes https://github.com/radicle-dev/drips-contracts/issues/242.


src/Drips.sol
```diff
@@ -337,17 +345,24 @@ abstract contract Drips {
         DripsHistory[] memory dripsHistory
     ) internal returns (uint128 amt) {
         uint256 squeezedNum;
-        uint256[] memory squeezedIdxs;
-        (amt, squeezedNum, squeezedIdxs) =
+        uint256[] memory squeezedRevIdxs;
+        bytes32[] memory historyHashes;
+        uint256 currCycleConfigs;
+        (amt, squeezedNum, squeezedRevIdxs, historyHashes, currCycleConfigs) =
```
xmxanuel, line 351: It is kind of interesting with how many different arrays we are dealing in these two functions. The complexity is really high.

1. `dripsHistory` array
   - includes all the information
2.  `squeezedHistoryHashes`
   - only the hashes from the provided dripsHistory
3. `nextSqueezed` array with` currCycleConfigs` as length
   - array can contain elements from outdated cycles :)
4. `squeezedRevIdxs` array with `squeezedNum` as length
   - basically how often we touched `nextSqueezed`
   - hashes in reverse order
5. `squeezedHistoryHashes` array for events length `squeezedNum`
   - array we construct only the present the hashes in the right order


Then it is important to note that we can have three different length
`squeezedNum != historyHashes.length != currCycleConfigs`.

(They can also be the same :) )

src/Drips.sol
```diff
@@ -359,34 +374,55 @@ abstract contract Drips {
...
-        squeezedIdxs = new uint256[](historyIdx);
-        while (squeezedIdx > 0 && historyIdx > 0) {
-            squeezedIdx--;
-            historyIdx--;
-            DripsHistory memory drips = dripsHistory[historyIdx];
+        for (uint256 i = 1; i <= dripsHistory.length && i <= currCycleConfigs; i++) {
+            DripsHistory memory drips = dripsHistory[dripsHistory.length - i];
             if (drips.receivers.length != 0) {
-                uint32 squeezeStart = nextSqueezed[squeezedIdx];
+                uint32 squeezeStart = nextSqueezed[currCycleConfigs - i];
```
xmxanuel, line 422: The name always confuses me a bit. Because the `squeezeStart` is only the squeeze start for the first squeeze in a new cycle in the other cases it is the `updateTime`.   It is kind of a `potentialsqueezeStart`

src/Drips.sol
```diff
@@ -359,34 +374,55 @@ abstract contract Drips {
...
                 if (squeezeStart < _currCycleStart()) squeezeStart = _currCycleStart();
                 if (squeezeStart < squeezeEnd) {
-                    squeezedIdxs[squeezedNum++] = squeezedIdx;
+                    squeezedRevIdxs[squeezedNum++] = i;
```
xmxanuel, line 425: It is a pity that the actual size of `squeezedRevIdxs` is unkown and we first need to iterate backwards at once to figure it out.

Outer-wise, we could have constructed it in the correct way.

xmxanuel: Maybe it make sense to to define the following variable

```solidity
historyHashIdx = currCycleConfigs - i;
```

```suggestion
                    squeezedRevIdxs[squeezedNum++] = historyHashIdx;
```

xmxanuel: Because the same subtraction happens in the higher function again

xmxanuel: Naming could be better because it is also the position in the `nextSqueezed[currCycleConfigs - i];`

xmxanuel (commented):
Overall, looks good. 

I think my main question is:
How important is the order of the emitted event hashes?
It would simplify the code if we can live with a reverse order.

src/Drips.sol
CodeSandwich: Yes, and that's terrible. It took me way too long trying to tie all these things together and weird `squeezedRevIdxs` was the best solution I've found :shrug:

CodeSandwich:
Emitting events in the reverse order would only simplify a single line:
```solidity
uint256 revIdx = squeezedRevIdxs[squeezedNum - i - 1];
```
would become:
```solidity
uint256 revIdx = squeezedRevIdxs[i];
```
I think that it's worth keeping, because it makes the API more elegant, more obvious and generally easier to handle.

src/Drips.sol
xmxanuel: Yes, it looks good. Just more an overall comment.

src/Drips.sol
xmxanuel: Ah true. You are right :+1:

xmxanuel (approved):

PR #245, opened 2022-12-12
Stop emitting unneeded Given events

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/243

### DIFF
--- a/src/Splits.sol
+++ b/src/Splits.sol
@@ -94,6 +94,10 @@ abstract contract Splits {
         _splitsStorageSlot = splitsStorageSlot;
     }
 
+    function _addSplittable(uint256 userId, uint256 assetId, uint128 amt) internal {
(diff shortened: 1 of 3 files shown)


src/Splits.sol
```diff
@@ -94,6 +94,10 @@ abstract contract Splits {
         _splitsStorageSlot = splitsStorageSlot;
     }
 
+    function _addSplittable(uint256 userId, uint256 assetId, uint128 amt) internal {
```
xmxanuel, line 97:
```suggestion
    function _addSplittable(uint256 receiver, uint256 assetId, uint128 amt) internal {
```
Maybe `receiver`?

xmxanuel (approved):
looks good

src/Splits.sol
xmxanuel: Okay. If we use a consistent pattern, I am fine.

PR #247, opened 2022-12-19
Refactor _calcMaxEnd

CodeSandwich:
A tiny change improving `_calcMaxEnd`:
- Better encapsulates handling of `config`
- Makes `_calcMaxEnd` the name of only one function


xmxanuel (approved):
looks good, it is cleaner

PR #248, opened 2022-12-20
Drop reserve

CodeSandwich:


src/DripsHub.sol
```diff
@@ -105,12 +105,11 @@ contract DripsHub is Managed, Drips, Splits {
...
     /// High value makes receiving cheaper by making it process less cycles for a given time range.
     /// Must be higher than 1.
-    /// @param reserve_ The address of the ERC-20 reserve which the drips hub will work with
-    constructor(uint32 cycleSecs_, IReserve reserve_)
+    constructor(uint32 cycleSecs_)
         Drips(cycleSecs_, erc1967Slot("eip1967.drips.storage"))
         Splits(erc1967Slot("eip1967.splits.storage"))
     {
-        reserve = reserve_;
+        return;
```
xmxanuel, line 112: Is it a compiler warning otherwise?

test/DripsHub.t.sol
```diff
@@ -378,16 +349,28 @@ contract DripsHubTest is Test {
...
     function balance() internal view returns (uint256) {
         return erc20.balanceOf(driver);
     }
 
     function assertBalance(uint256 expected) internal {
-        assertEq(erc20.balanceOf(driver), expected, "Invalid balance");
+        assertEq(balance(), expected, "Invalid balance");
+    }
+
+    function dripsHubBalance() internal view returns (uint256) {
```
xmxanuel, line 368: In general good design of all the tests. That didn't result in 100 changes of replacing `balanceOf` reserve with `balanceOf` DripsHub.

src/Reserve.sol
```diff
@@ -1,258 +0,0 @@
...
-    /// If used recklessly, it may cause a mismatch between the counter and the actual balance
-    /// making valid future calls to `withdraw` or `setPlugin` fail due to lack of funds.
-    /// Callable only by the current owner.
-    /// @param token The used token.
-    /// @param plugin The plugin to withdraw from.
-    /// It doesn't need to be registered as a plugin for `token`.
-    /// Pass `NO_PLUGIN` to withdraw directly from the reserve balance.
-    /// @param to The address to which funds are withdrawn.
-    /// @param amt The withdrawn amount.
-    function forceWithdraw(IERC20 token, IReservePlugin plugin, address to, uint256 amt)
```
xmxanuel, line 210: So, we decided to completely remove the `forceWithdraw` functionality?

xmxanuel (approved):
nice work. :100:   Always good to remove more code than to add more of it.

src/Reserve.sol
CodeSandwich: Yes, it was needed first and foremost for collecting income from plugins by shaving off tokens without touching the internal counter. For extreme cases like emergency withdrawal after a total collapse of the protocol we'll need to update the contract and change its logic to do some kind of a refunding or at least draining to a given address. Do you think that we should have the draining function available at all times for the owner to call?

src/DripsHub.sol
CodeSandwich: It triggers `solhint`'s empty block warning.

PR #250, opened 2022-12-21
Add tips docs

CodeSandwich:
It's the same docs string in 4 files.

### DIFF
--- a/src/AddressDriver.sol
+++ b/src/AddressDriver.sol
@@ -94,6 +94,29 @@ contract AddressDriver is Upgradeable, ERC2771Context {
     /// Positive to add funds to the drips balance, negative to remove them.
     /// @param newReceivers The list of the drips receivers of the sender to be set.
     /// Must be sorted by the receivers' addresses, deduplicated and without 0 amtPerSecs.
+    /// @param maxEndTip1 An optional parameter allowing gas optimization, pass `0` to ignore it.
+    /// The first tip for finding the maximum end time when all drips stop due to funds
+    /// running out after the balance is updated and the new receivers list is applied.
+    /// Tips have no effect on the results of calling this function, except potentially saving gas.
+    /// Tips are Unix timestamps used as the starting points for binary search for the time
+    /// when funds run out in the range of timestamps from the current block's to `2^32`.
+    /// Tips lower than the current timestamp are ignored.
+    /// You can provide zero, one or two tips. The order of tips doesn't matter.
+    /// Tips are the most effective when one of them is lower than or equal to
+    /// the time when funds run out, and the other one is strictly larger than that time,
+    /// the smaller the difference between such tips, the higher gas savings.
+    /// The savings are the highest possible when one of the tips
+    /// is equal to the time when funds run out and the other one is larger by 1.
+    /// It's worth noting that the exact timestamp of the block in which this function is executed
+    /// may affect correctness of the tips, especially if they're precise.
+    /// Tips don't provide any benefits when balance is not enough to cover
+    /// a single second of dripping or is enough to cover all drips until timestamp `2^32`.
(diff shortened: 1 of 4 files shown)


src/AddressDriver.sol
```diff
@@ -94,6 +94,27 @@ contract AddressDriver is Upgradeable, ERC2771Context {
     /// Positive to add funds to the drips balance, negative to remove them.
     /// @param newReceivers The list of the drips receivers of the sender to be set.
     /// Must be sorted by the receivers' addresses, deduplicated and without 0 amtPerSecs.
+    /// @param maxEndTip1 The first tip to calculate the maximum end time of drips
```
xmxanuel, line 97:
```suggestion
    /// @param maxEndTip1 optional parameter for gas optimization pass zero if it should not be used
```
I would start with a one-liner for readers who are not interested in using the parameter and are not interested in what it is exactly.  I just want them to read the first line and get all the information they need.

src/AddressDriver.sol
```diff
@@ -94,6 +94,27 @@ contract AddressDriver is Upgradeable, ERC2771Context {
...
+    /// is equal to the maximum end time and the other one is larger by 1.
+    /// It's worth noting that the exact timestamp of the block in which this function is executed
+    /// may affect correctness of the tips, especially if they're precise.
+    /// Tips are used to divide the binary searched range of future timestamps into smaller parts,
+    /// so even inaccurate tips can be useful, and providing a single tip
+    /// or two tips that don't enclose the maximum end time can still save some gas.
+    /// Providing tips that don't limit the searched time range significantly
+    /// may cause slightly higher gas usage than not providing any tips.
+    /// @param maxEndTip2 The second tip to calculate the maximum end time of drips
+    /// after the balance and the receivers list are updated.
```
xmxanuel, line 117: Nice explanation. I would the information that the tips are acutally Unix timestamps.  Which is never said directly.

xmxanuel: Out of curiosity and because I would have called them `hints` instead of `tips`. 

> A tip is when you tell a person useful information, to help them. A hint is when you don't actually tell them, you say something different which might help them find that information themselves. 

For me it would be more a `hint` but I guess both is fine

src/AddressDriver.sol
```diff
@@ -94,6 +94,27 @@ contract AddressDriver is Upgradeable, ERC2771Context {
...
+    /// and passing a zero is an easy way to not provide a tip.
+    /// It's acceptable to provide zero, one or two tips. The order of tips doesn't matter.
+    /// Tips are the most effective when one of them is lower than or equal to the maximum end time,
+    /// and the other one is strictly larger than the maximum end time,
+    /// the smaller the difference between them, the higher gas savings.
+    /// The savings are the highest possible when one of the tips
+    /// is equal to the maximum end time and the other one is larger by 1.
+    /// It's worth noting that the exact timestamp of the block in which this function is executed
+    /// may affect correctness of the tips, especially if they're precise.
+    /// Tips are used to divide the binary searched range of future timestamps into smaller parts,
```
xmxanuel, line 111:
```suggestion
    /// Tips are used to find the maxEnd (running out of funds) time with a binary search by dividing the search range of future timestamps into smaller parts,
```

Maybe?

src/AddressDriver.sol
CodeSandwich: Good point, added.

src/AddressDriver.sol
CodeSandwich: Added a note about Unix timestamps. Renaming tips to hints is a good idea, will apply in the next PR.

src/AddressDriver.sol
CodeSandwich: I agree, that's clearer, applied something similar.

xmxanuel (approved):

PR #251, opened 2022-12-28
Add pauser role

### LINKED ISSUE #249: Add "pauser" role
As of now in order to pause `DripsHub` one needs to have the ownership privilages. This mostly defeats the purpose of contract's pausability, because the owner can just update the protocol implementation address to a contract not doing anything to pause and back to the original to unpause. This is also problematic in terms of governance, because being able to change the entire protocol is the highest existing privilege, which should be well guarded with a preferably slow process requiring consent from many parties. On the other hand pausing is less harmful and in case of an emergency should be quick to activate with less overhead. Some protocols even choose to automate pausing with blockchain monitoring software to minimize reaction time.

A solution I propose here is to add a "pauser" role, which allows pausing and unpausing. Unlike ownership there can be multiple addresses with this role, granting and revocation of this role is done by the owner. The owner can pause and unpause without being a pauser.

CodeSandwich:
Closes #249. Adds the pauser role for DripsHub. The `optimizer_runs` reduction is necessary because the contract was becoming too large to deploy, the performance impact is consistently less than 0.5% for Foundry test runs.

### DIFF
--- a/src/Managed.sol
+++ b/src/Managed.sol
@@ -3,6 +3,9 @@ pragma solidity ^0.8.17;
 
 import {Upgradeable} from "./Upgradeable.sol";
 import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
+import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
+
+using EnumerableSet for EnumerableSet.AddressSet;
 
 /// @notice A mix-in for contract pauseability and upgradeability.
 /// It can't be used directly, only via a proxy. It uses the upgrade-safe ERC-1967 storage scheme.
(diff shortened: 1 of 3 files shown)


src/Managed.sol
```diff
@@ -3,23 +3,49 @@ pragma solidity ^0.8.17;
 
 import {Upgradeable} from "./Upgradeable.sol";
 import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
+import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
+
+using EnumerableSet for EnumerableSet.AddressSet;
```
xmxanuel, line 8: What is the mean reason for the EnumerableSet?  Only to list all pausers with `allPausers()` I assume.

I think it is okay to use it, maybe a bit overengineered. Since, we rely on so many things just on events. 

However, we could argue that these roles come with great power, so it seems good. :+1:

src/Managed.sol
```diff
@@ -3,23 +3,49 @@ pragma solidity ^0.8.17;
...
 /// @notice A mix-in for contract pauseability and upgradeability.
 /// It can't be used directly, only via a proxy. It uses the upgrade-safe ERC-1967 storage scheme.
 ///
 /// All instances of the contracts are paused and can't be unpaused.
 /// When a proxy uses such contract via delegation, it's initially unpaused.
 abstract contract Managed is Upgradeable {
-    /// @notice The pointer to the storage slot with the boolean holding the paused state.
-    bytes32 private immutable pausedSlot = erc1967Slot("eip1967.managed.paused");
+    /// @notice The pointer to the storage slot holding a single `ManagedStorage` structure.
+    bytes32 private immutable _storageSlot = erc1967Slot("eip1967.managed.storage");
```
xmxanuel, line 17:
```suggestion
    bytes32 private immutable _storageSlot = _erc1967Slot("eip1967.managed.storage");
```
I just realized this, if we want to be consistent we should change it.

xmxanuel (approved):
looks good.

src/Managed.sol
CodeSandwich: Yes, that's the reason. With `allPausers` you can even see on-chain who's a pauser, and it's trivial to query for them off-chain, it's a press of a button on Etherscan. When relying on events, it's more difficult and you don't get this information on-chain. The gas cost is larger, because instead of updating 1 slot you need to update 3, but IMO that's a fair trade off, pausers are not supposed to be updated often.

src/Managed.sol
CodeSandwich: Good catch, will fix it in an upcoming PR!

PR #252, opened 2022-12-31
Add fuzz test for splits adding up

CodeSandwich:


test/Splits.t.sol
```diff
@@ -298,4 +298,62 @@ contract SplitsTest is Test, Splits {
...
+            for (uint256 j = i + 1; j < receiversRaw.length; j++) {
+                if (receiversRaw[i].userId > receiversRaw[j].userId) {
+                    (receiversRaw[i], receiversRaw[j]) = (receiversRaw[j], receiversRaw[i]);
+                }
+            }
+        }
+        uint256 unique = 0;
+        for (uint256 i = 0; i < receiversRaw.length; i++) {
+            if (receiversRaw[i].userId != receiversRaw[unique].userId) unique++;
+            receiversRaw[unique] = receiversRaw[i];
```
xmxanuel, line 317: If the idea here is to find out, how many unique addresses are in the `receiverRaw` list. 

I think this algorithm won't work all the time correctly.

Counterexample:

```
// example user-ids
[2,2,2,2,2,2,2,2,2,3,3,3,3,3,3,3,3,3]
```

xmxanuel: However, it is okay from the DripsHub perspective to have the same userID, multiple times if correctly sorted.

CodeSandwich: The goal is not only to count unique user IDs, but also to remove duplicates. The result from your example will be `[2,3]`, which is a valid list of splits receivers. This is important, because splits receivers lists DO NOT accept duplicate user IDs. We can either drop the duplicates (easy, we do it here), or merge them, e.g. by adding up their weights (difficult).

xmxanuel:
```solidity
    function testUnique() public {
    receiversRaw = [2,2,2,2,2,2,2,2,2,3,3,3,3,3,3,3,3,3];
    uint256 unique = 0;
        for (uint256 i = 0; i < receiversRaw.length; i++) {
            if (receiversRaw[i] != receiversRaw[unique]) unique++;
        }
        console.log(unique);
    }
```

```
Running 1 test for test/Counter.t.sol:CounterTest
[PASS] testUnique() (gas: 440117)
Logs:
  9

```

xmxanuel: > This is important, because splits receivers lists DO NOT accept duplicate user IDs.
Ahh, but we accept it for `setDrips`.

CodeSandwich: There's a part missing from your example code, it's shifting of the unique entries:
```solidity
function testUnique() public {
    uint8[18] memory receiversRaw = [2,2,2,2,2,2,2,2,2,3,3,3,3,3,3,3,3,3];
    uint256 unique = 0;
    for (uint256 i = 0; i < receiversRaw.length; i++) {
        if (receiversRaw[i] != receiversRaw[unique]) unique++;
        receiversRaw[unique] = receiversRaw[i];
    }
    emit log_uint(unique);
    emit log_uint(receiversRaw[0]);
    emit log_uint(receiversRaw[1]);
}
```
```
Running 1 test for test/Splits.t.sol:SplitsTest
[PASS] testUnique() (gas: 9910)
Logs:
  1
  2
  3
```

xmxanuel: Ah, you are right!  Then it works. :+1:

xmxanuel: A lot of copying in case of duplicates but it is just a test.

PR #255, opened 2023-01-03
Make drivers pauseable

CodeSandwich:


src/NFTDriver.sol
```diff
@@ -223,11 +229,54 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Upgradeable {
...
+        super.safeTransferFrom(from, to, tokenId, data);
+    }
+
+    /// @inheritdoc IERC721
+    function setApprovalForAll(address operator, bool approved) public override whenNotPaused {
+        super.setApprovalForAll(operator, approved);
+    }
+
+    /// @inheritdoc IERC721
+    function transferFrom(address from, address to, uint256 tokenId)
```
xmxanuel, line 272: This should be up for debate. I understand the pausable for the drips-related function like collect, setDrips etc.
However, let's say a user's security setup is not good enough anymore (compromised keys etc). 

Why should we block transferring the ownership?  We anyway assume OZ is safe.

test/NFTDriver.t.sol
```diff
@@ -253,4 +253,63 @@ contract NFTDriverTest is Test {
 
         assertEq(dripsHub.splittable(tokenId, erc20), amt, "Invalid splittable after give");
     }
+
+    modifier canBePausedTest() {
```
xmxanuel, line 257: nice way to test it.

src/NFTDriver.sol
CodeSandwich: That's a good questionThat's a good question. My assumption was that if the NFTDriver contract is broken, all its interactions with the outer world and its entire state should be frozen, no exceptions. Half-measures like allowing token transfers on a broken contract may lead to deepening issues in the storage. Any emergency state restorations will be harder if the users are allowed to change it over the whole course of maintenance with no way to stop them. The users may also harm themselves if their NFT identities temporarily work incorrectly or don't work at all, but they can still be transferred, traded or burned. Later, when the contract is fixed and unpaused, the users may find themselves locked out of perfectly fine identities. It's probably better to clearly signal a failure than to hope that oblivious users don't hurt themselves. The OZ NFT implementation may be safe, but it's just a part of the bigger machine.

xmxanuel: It is anyway an edge case and it might depend on the underlying issue. However, I agree with you that a general stop might be safer.

PR #257, opened 2023-01-04
Merge Upgradeable into Managed

CodeSandwich:
Keeping `Upgradeable` and `Managed` as separate blocks built on top of one another doesn't seem to make sense anymore. Pauseability is a great addition to upgradeability as two supplementing parts of emergency management, and none of them is powerful enough without the other. I don't think that in a foreseeable future we should have upgradeable, but not pauseable contracts, and the recent addition of pauseability to all upgradeable contracts confirm that.


xmxanuel (approved):
looks good.

PR #263, opened 2023-01-10
Update README.md

CodeSandwich:
Closes https://github.com/radicle-dev/drips-contracts/issues/77.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
-# Radicle Drips Hub Contracts
+# Radicle Drips protocol V2 smart contracts
 
-Radicle Drips Hub is the smart contract running the drips and splits ecosystem.
+Radicle Drips is an EVM blockchain protocol for streaming and splitting ERC-20 tokens.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,30 +12,30 @@ Then, run the following command to install the dependencies:
...
-### Run specific tests
-A regular expression can be used to only run specific tests.
+## Run tests with extra options
+`forge test` can be used directly to provide extra options, like running only specific tests.
 
 ```bash
-forge test -m <REGEX>
-forge test -m testName
-forge test -m ':ContractName\.'
+forge test --mt testName
```
xmxanuel, line 34: Just learned about match test instead of match. What is the difference?

xmxanuel (approved):

README.md
CodeSandwich: There's none. `-m` is a legacy flag from the DappTools era, and it's deprecated in favor of `--match-test` or `--mt`. This is because Foundry has new flags for matching based on contract name and path name, plain `-m` was too vague.

PR #268, opened 2023-01-16
Drop docs for how the pool works

CodeSandwich:
The old docs are redundant, outdated and as of now incomplete, devs should read https://v2.docs.drips.network/docs/the-protocol/technical-overview.


xmxanuel (approved):
:+1:

PR #272, opened 2023-02-06
Fix squeezing drips with start in the past

CodeSandwich:


xmxanuel (approved):
looks good.

I started to wonder if there was another case we might also miss. So, far I couldn't find one.

PR #274, opened 2023-02-06
Fix NFTDriver token metadata

CodeSandwich:
OZ stores name and symbol in storage initialized in constructor. Putting the token contract behind a proxy replaces storage with an empty one clearing the token metadata.

### DIFF
--- a/src/NFTDriver.sol
+++ b/src/NFTDriver.sol
@@ -9,7 +9,8 @@ import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 import {
     ERC721,
     ERC721Burnable,
-    IERC721
+    IERC721,
+    IERC721Metadata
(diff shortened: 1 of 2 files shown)


xmxanuel (approved):

src/NFTDriver.sol
```diff
@@ -240,6 +241,16 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Managed {
         dripsHub.emitUserMetadata(tokenId, userMetadata);
     }
 
+    /// @inheritdoc IERC721Metadata
+    function name() public pure override returns (string memory) {
```
xmxanuel, line 245: I just tried if you could `override` just with a constant variable. Doesnt' work :)

CodeSandwich: I've tried it too, `name` would need to be `external`. It kind of makes sense, for internal use a function and an immutable are two very differently operated entities, which in case of `override` would have the same name.

PR #275, opened 2023-02-07
Minor cleanups

CodeSandwich:


src/DripsHub.sol
```diff
@@ -68,6 +68,10 @@ contract DripsHub is Managed, Drips, Splits {
     /// @notice The total amount the contract can store of each token.
     /// It's the minimum of _MAX_TOTAL_DRIPS_BALANCE and _MAX_TOTAL_SPLITS_BALANCE.
     uint256 public constant MAX_TOTAL_BALANCE = _MAX_TOTAL_DRIPS_BALANCE;
+    /// @notice On every timestamp `T`, which is a multiple of `cycleSecs`, the receivers
+    /// gain access to drips received during `T - cycleSecs` to `T - 1`.
+    /// Always higher than 1.
```
xmxanuel, line 73: > /// Always higher than 1.
Mhm, but I could still set it to `1`, no? We are not checking this.

xmxanuel (approved):

src/DripsHub.sol
CodeSandwich: We are, `require(cycleSecs > 1, "Cycle length too low");`

PR #276, opened 2023-02-07
Add minAmtPerSec

CodeSandwich:
Ban amtPerSec so low that nothing is dripped.

### DIFF
--- a/src/Drips.sol
+++ b/src/Drips.sol
@@ -116,6 +116,9 @@ abstract contract Drips {
     /// Always higher than 1.
     // slither-disable-next-line naming-convention
     uint32 internal immutable _cycleSecs;
+    /// @notice The minimum amtPerSec of a drip. It's 1 token per cycle.
+    // slither-disable-next-line naming-convention
+    uint160 internal immutable _minAmtPerSec;
     /// @notice The storage slot holding a single `DripsStorage` structure.
     bytes32 private immutable _dripsStorageSlot;
 
@@ -220,6 +223,7 @@ abstract contract Drips {
     constructor(uint32 cycleSecs, bytes32 dripsStorageSlot) {
         require(cycleSecs > 1, "Cycle length too low");
         _cycleSecs = cycleSecs;
+        _minAmtPerSec = (_AMT_PER_SEC_MULTIPLIER + cycleSecs - 1) / cycleSecs;
         _dripsStorageSlot = dripsStorageSlot;
     }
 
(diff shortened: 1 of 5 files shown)


src/Drips.sol
```diff
@@ -220,6 +223,7 @@ abstract contract Drips {
     constructor(uint32 cycleSecs, bytes32 dripsStorageSlot) {
         require(cycleSecs > 1, "Cycle length too low");
         _cycleSecs = cycleSecs;
+        _minAmtPerSec = (_AMT_PER_SEC_MULTIPLIER + cycleSecs - 1) / cycleSecs;
```
xmxanuel, line 226: What is the reasoning here: 

My indutive way would be:
```
// equation:  solve for minAmtPerSec
minAmtPerSec * cycleSecs/_AMT_PER_SEC_MULTIPLIER = 1

minAmtPerSec = _AMT_PER_SEC_MULTIPLIER/cycleSecs
```

Why the `cycleSecs - 1` part in your `minAmtPerSec` definition?

xmxanuel: I changed it to my formula and wrote a small test that passes. I guess the `cycleSecs - 1` is for some edge case?

```solidity
    function testMinAmtPerSec() public {
        DripsReceiver[] memory receivers = new DripsReceiver[](1);
        uint duration = 0;
        uint start = 0;
        uint cycleSecs = 10;
       uint256 amtPerSecFull = Drips._AMT_PER_SEC_MULTIPLIER/cycleSecs;
        DripsConfig config = DripsConfigImpl.create(
            uint32(receiver), uint160(amtPerSecFull), uint32(start), uint32(duration)
        );
        receivers[0] = DripsReceiver(receiver, config);
        // deactivated expectedMaxEndFromNow check in setDrips testhelper
        setDrips(sender, 0, 100, receivers, 50);
        skip(10);
        receiveDrips(receiver, 1);
    }
```

CodeSandwich: I've changed it to your formula and now the existing `testMinAmtPerSec` is failing. The reason for `+ cycleSecs - 1` is that it rounds the division up, we don't want to drip per cycle **almost** 1 token, we want **at least** 1 token. Your test is passing because `cycleSecs = 10` coincidentally cleanly divides `_AMT_PER_SEC_MULIPLIER = 10^9`.

xmxanuel: Ah true. You are right. If cycleSecs would be `3` it would not be correct.  :+1: 
Would be nice to add a `divRoundUp` helper, this would make it clearer.

xmxanuel (approved):

PR #279, opened 2023-02-17
C4 small improvements

CodeSandwich:
A handful of small improvements, segregated into docs and code.

### DIFF
--- a/src/AddressDriver.sol
+++ b/src/AddressDriver.sol
@@ -15,7 +15,7 @@ import {SafeERC20} from "openzeppelin-contracts/token/ERC20/utils/SafeERC20.sol"
 
 /// @notice A DripsHub driver implementing address-based user identification.
 /// Each address can use `AddressDriver` to control a single user ID derived from that address.
-/// No registration is required, an `AddressDriver`-based user ID for each address is know upfront.
+/// No registration is required, an `AddressDriver`-based user ID for each address is known upfront.
 contract AddressDriver is Managed, ERC2771Context {
     using SafeERC20 for IERC20;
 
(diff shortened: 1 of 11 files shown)


src/AddressDriver.sol
```diff
@@ -43,47 +42,47 @@ contract AddressDriver is Managed, ERC2771Context {
 
     /// @notice Calculates the user ID for the message sender
     /// @return userId The user ID
-    function callerUserId() internal view returns (uint256 userId) {
+    function _callerUserId() internal view returns (uint256 userId) {
```
xmxanuel, line 45: :+1:

src/Caller.sol
```diff
@@ -204,6 +204,6 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
         returns (bytes memory returnData)
     {
         // Encode the message sender as per ERC-2771
-        return Address.functionCallWithValue(to, abi.encodePacked(data, sender), value);
+        return Address.functionCallWithValue(to, bytes.concat(data, bytes20(sender)), value);
```
xmxanuel, line 207: interesting. Is the concat cheaper?
Why the bytes20 cast also cheaper?

src/Drips.sol
```diff
@@ -288,11 +288,13 @@ abstract contract Drips {
             receivableCycles = toCycle - fromCycle - maxCycles;
             toCycle -= receivableCycles;
         }
-        DripsState storage state = _dripsStorage().states[assetId][userId];
+        mapping(uint32 => AmtDelta) storage amtDeltas =
+            _dripsStorage().states[assetId][userId].amtDeltas;
         for (uint32 cycle = fromCycle; cycle < toCycle; cycle++) {
-            amtPerCycle += state.amtDeltas[cycle].thisCycle;
+            AmtDelta memory amtDelta = amtDeltas[cycle];
```
xmxanuel, line 294: This makes sense

xmxanuel (approved):
looks good

src/Caller.sol
CodeSandwich: No, it's the same, it's just that `abi.encodePacked` is now deprecated in Solidity in favor of `bytes.concat`.

src/Drips.sol
CodeSandwich: It actually saves a `sload`, I was a little surprised that it's not optimized better. Not all tips from C4 actually make a difference when benched, but this one does.

PR #280, opened 2023-03-01
Transfer tokens directly to DripsHub

CodeSandwich:


xmxanuel (commented):
I don't really like the pattern. For me, it is still additional complexity and a not-so-common pattern. 

A public `withdraw` function together with these increase/decrease. In the Driver we would still use the `transferFromCaller` pattern.

Basically, two different ways of taking funds in the system. 

An EOA as driver would be an anti-pattern, however, we don't block it. 
For me just saving one ERC20 transferFrom is not worth it.

src/AddressDriver.sol
```diff
@@ -58,7 +58,7 @@ contract AddressDriver is Managed, ERC2771Context {
     /// @return amt The collected amount
     function collect(IERC20 erc20, address transferTo) public whenNotPaused returns (uint128 amt) {
         amt = dripsHub.collect(_callerUserId(), erc20);
-        erc20.safeTransfer(transferTo, amt);
+        if (amt > 0) dripsHub.withdraw(erc20, transferTo, amt);
```
xmxanuel, line 61: Are we planning to implement a toAddress for the collect? 

It makes writing a correct Driver just harder.
Like if you call `collect` don't forget to call `withdraw` as well afterward otherwise your funds are gone.

CodeSandwich:
Thanks for the input @xmxanuel! I can't argue with the general feeling, but I'll try to lay out some counter arguments.
- The pattern isn't anything exotic. It's implemented in OZ for Ether as [PullPayment](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/PullPayment.sol), there are some links in the docs about why it's safer than push-sending funds from inside of the contract logic. ERC-20 must be handled just as carefully as raw Ether because any operation is giving away the control flow, e.g. there are ERC-777 hooks. Another example is Uniswap where `burn`ing your position doesn't transfer your tokens out, you need to call `collect` separately in [UniswapV3Pool](https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L496:L548). A similar pattern is used when `mint`ing, the Pool doesn't `transferFrom` anybody, it expects tokens to be transferred in by the periphery, which handles `transferFrom`. [UniswapV3Pool](https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L488) calls [LiquidityManagement](https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/LiquidityManagement.sol#L25) which does `transferFrom` in [PeripheryPayments](https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/PeripheryPayments.sol#L67). It's not identical as in DripsHub, it's tailored for a different protocol used differently, but the pattern is there, the core of the protocol doesn't handle transfers.
- Approvals are a huge pain to handle generically, there are different implementations and weird ways to handle it, leaving it to the 3rd parties, ideally EOAs, is by far the safest startegy from our side.
- With the proposed changes DripsHub API becomes reentrancy safe. Any 3rd party calls are limited to `balanceOf` which is `view`, so it's much easier to protect the driver from reentrancy. Only `withdraw` must be considered a threat, but it's an explicit step that can be easily made after cleaning up the state.
- The current and new designs are equivalent when it comes to difficulty of handling the funds taken out of DripsHub e.g. when collecting. In the new design you need to remember to explicitly call `withdraw(erc20, user, amt)` to give the funds to the user, but in the current design you need to remember to call `erc20.transfer(user, amt)`. It's exactly the same bug opportunity.
- The EOA driver is indeed an anti-pattern, that's why it's not worth paying for an extra ERC-20 transfer on each operation to preserve it. According to Etherscan a basic transfer of generic ERC-20 tokens on Mainnet currently costs about 4 USD in gas, we're talking about real money savings here. Opening door to tricks like performing operations in DripsHub without a single ERC-20 transfer is also a nice direction to explore in the future.

xmxanuel:
My main point is not about withdraw push vs pull. A pull withdraw is a common pattern.

I never said it is exotic. It is just a bit more complex to understand. Wouldn't you agree?

It is about topUp with push erc20  and the topUp call needs to happen in the same transaction otherwise someone else might take it. I know the pattern for FlashSwaps but here we have a callback pattern that needs to perform the erc20.push and the check happens after the callback call. 

I think I wouldn't be that hesitant if it is just within Drip's contract but it is an external entry point callable by anyone they just need to register a driver ID first.  

I just looked at your first reference. This is a pull payment based on the stored amount in the escrow. Not really comparable in my opinion. 

I will check out the other references to get deeper into the topic.

I basically want to see an external contract function more or less callable by anyone who compares the erc20 balance with some stored balances to topUp.

xmxanuel:
I can definitely see the gas benefit for our existing Drivers. This is definitely a strong argument.

xmxanuel:
> Approvals are a huge pain to handle generically, there are different implementations and weird ways to handle it, leaving it to the 3rd parties, ideally EOAs, is by far the safest strategy from our side.

I can see your point.

It is still a hot take and a bit against the philosophy of Ethereum where both EOA and contracts are first-class citizens. 

Which we would break with this PR. 

We basically say to use one of our existing driver's to maintain multiple accounts with one address or write your own driver.  Which will be in the end more expensive instead of direct DripsHub calls :)

CodeSandwich:
I wouldn't worry too much about EOAs, we're discussing an internal API between the core contract and the periphery contracts. It's normal for such APIs to be based on callbacks and the ability to perform multiple actions in a single transaction, they aren't supposed to be run by EOAs. The EOAs and generally end users should use the periphery contracts and not bother with the internal APIs. Being able to use an EOA as a driver was a weird side effect without a clear use case. If we really need to let a single EOA address control 2^224 user IDs, we can write a simple driver with a cloning factory for that.

xmxanuel:
I believe there is some ambiguity regarding whether this API is considered public or internal.
It is definitely not a full internal API. If you think so, we should restrict the register driver function. 

However, you are right that most developers and end users won't interact with it. 

My main concern, however, was the additional complexity. Despite this, we have already made significant changes to the storage/hashes to ensure gas-efficient calls, and these changes are acceptable. (In terms of complexity added).

Therefore, if we both think it is safe enough from a security perspective we can add it.

In general, I like your efforts to make it even more gas efficient and optimize the last remaining parts. :muscle: 

To address any confusion, I suggest we add some comments to the `registerDriver` function to provide clarity about EOA.

CodeSandwich:
I've added a comment clarifying the requirements for the drivers.

IMO the proposed changes actually improve the security, it's a clearer design when it comes to interacting with 3rd parties. WDYT?

Do you have any additional comments about the design or the code itself?

xmxanuel (approved):
looks good

PR #281, opened 2023-03-01
Ban registering drivers for 0 address

CodeSandwich:
This makes the theoretical attack of draining all driver IDs 4 times slower. Plus it never made sense to register drivers on address `0`.


CodeSandwich:
It's probably fine the way it is. The cost to register a driver is already quite high, it'd require the entire blocks capacity of Ethereum from 120 days or Polygon from 20 days to drain all the driver IDs. Increasing it 4-fold makes it even less feasible.

xmxanuel:
Is this really necessary? It is fine for me to merge it. 
You are right with your reasoning but you could say the same for address 0x01 and 0x2 ...

CodeSandwich:
I agree, it's a weak prevention of bugs, but most importantly it's a DoS slow down, because address `0` is much cheaper to store.

xmxanuel:
Yeah, lets merge it.

PR #285, opened 2023-03-03
Add unauthorizeAll to Caller

CodeSandwich:
It required some storage structure changes and an extra SLOAD when making any authorization related operations. The strategy is to forget the existing set of authorized addresses and just start using a new clean instance. Clearing an existing set would require few SSTORE operations per unauthorized address, so it could be easily made uncallable by authorizing just a few thousand dummy addresses.

### DIFF
--- a/src/Caller.sol
+++ b/src/Caller.sol
@@ -85,9 +85,18 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
 
     /// @notice Each sender's set of address authorized to make calls on its behalf.
     // slither-disable-next-line naming-convention
-    mapping(address => EnumerableSet.AddressSet) internal _authorized;
+    mapping(address => AddressSetClearable) internal _authorized;
     /// @notice The nonce which needs to be used in the next EIP-712 message signed by the address.
     mapping(address => uint256) public nonce;
+
+    /// @notice A clearable set of addresses.
+    /// @param clears Number of performed clears. Increase to clear.
+    /// @param addressSets The key is the number of performed clears and the value is the set.
+    /// Always use the set under the key equal to the current value of `clears`.
+    struct AddressSetClearable {
+        uint256 clears;
+        mapping(uint256 => EnumerableSet.AddressSet) addressSets;
+    }
 
     /// @notice Emitted when granting the authorization
     /// of an address to make calls on behalf of the `sender`.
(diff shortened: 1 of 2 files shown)


src/Caller.sol
```diff
@@ -85,10 +85,19 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
...
-    mapping(address => EnumerableSet.AddressSet) internal _authorized;
+    mapping(address => AddressSetClearable) internal _authorized;
     /// @notice The nonce which needs to be used in the next EIP-712 message signed by the address.
     mapping(address => uint256) public nonce;
 
+    /// @notice A clearable set of addresses.
+    /// @param clears Number of performed clears. Increase to clear.
+    /// @param addressSets The key is the number of performed clears and the value is the set.
+    /// Always use the set under the key equal to the current value of `clears`.
+    struct AddressSetClearable {
```
xmxanuel, line 96: You are completely right that it would be very expensive to remove. :+1: 

Very nice idea with a clear `counter` and new mapping slot for the addresses :+1:

src/Caller.sol
```diff
@@ -101,36 +110,47 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
...
     /// @param user The unauthorized address.
     function unauthorize(address user) public {
         address sender = _msgSender();
-        require(_authorized[sender].remove(user), "Address is not authorized");
+        require(_getAuthorizedSet(sender).remove(user), "Address is not authorized");
         emit Unauthorized(sender, user);
     }
 
+    /// @notice Revokes all authorizations to make calls on behalf of the sender.
+    function unauthorizeAll() public {
```
xmxanuel, line 134: Each authorized address could `unauthorizeAll` all other authorized addresses. (including itself).

xmxanuel: A scenario could be that I don't like a specific action another authorized address wants to perform. 
I front-run it with an `unauthorizeAll` call to delay the action.

src/Caller.sol
```diff
@@ -101,36 +110,47 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
...
+        address sender = _msgSender();
```
xmxanuel, line 135:
```suggestion
        address sender = msg.sender;
        require(msg.sender != address(this));
```
What would we lose if we restrict it to the actual sender?

xmxanuel (commented):
Very nice design.

Some questions if we should restrict the power of unauthorizeAll?

src/Caller.sol
CodeSandwich: Yes, cleaning all authorizations while being authorized removes your own authorization too, that's probably the expected thing that should happen. You can explicitly add an exception by batching `unauthorizeAll` with a bunch of `authorize`s, e.g. re-authorizing yourself. 

Yes, all of this is correct. It's the same as with `unauthorize`, each authorized address may unauthorize the others, `unauthorizeAll` is just an optimization.

xmxanuel: Yes, but why do we think is it better than keeping this method only callable by the actual owner?

src/Caller.sol
CodeSandwich: We'd lose the flexibility, like being able to batch or call via an operator. It wouldn't change much with the security model, because it's still possible to call `unauthorize` in a loop, it's just more expensive.

xmxanuel: Yeah, for the `unauthorize` this make sense but for  `unauthorizeAll` ?

CodeSandwich: It's the same for both of them, they are perfectly equivalent to each other. If you restrict either one of them, the other can be used as a less gas efficient substitute.

xmxanuel: Yeah, fair point. Let's keep it :+1:

xmxanuel (approved):
looks good

PR #286, opened 2023-03-03
Add more Caller events

CodeSandwich:
It's probably a good thing to at least signal making calls on behalf of others. Plus the current nonce now can be taken straight from events.


xmxanuel (approved):
looks good

PR #287, opened 2023-03-03
Add setNonce to Caller

CodeSandwich:
The maximum nonce increase should be sufficient for all practical purposes, but it solves the risk of ever draining nonces.

### DIFF
--- a/src/Caller.sol
+++ b/src/Caller.sol
@@ -79,6 +79,8 @@ struct Call {
 /// It also applies to `authorize` and `unauthorize`, they too can be called using
 /// `callAs`, `callSigned` or `callBatched`.
 contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
+    /// @notice The maximum increase of the nonce possible by calling `setNonce`.
+    uint256 public constant MAX_NONCE_INCREASE = 10 ** 9;
     string internal constant CALL_SIGNED_TYPE_NAME = "CallSigned("
         "address sender,address target,bytes data,uint256 value,uint256 nonce,uint256 deadline)";
     bytes32 internal immutable callSignedTypeHash = keccak256(bytes(CALL_SIGNED_TYPE_NAME));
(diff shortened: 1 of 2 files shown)


src/Caller.sol
```diff
@@ -215,6 +222,18 @@ contract Caller is EIP712("Caller", "1"), ERC2771Context(address(this)) {
         return _call(sender, target, data, msg.value);
     }
 
+    /// @notice Sets the new nonce for the sender.
+    /// @param newNonce The new nonce.
+    /// It must be larger than the current nonce but by no more than MAX_NONCE_INCREASE.
+    function setNonce(uint256 newNonce) public {
+        address sender = _msgSender();
```
xmxanuel, line 229: Currently, with the `approve` pattern an authorized address would be able to call `setNonce` as well and basically prevent the actual owner from using the Caller.

Again, we already had all these trust discussions but I wanted to point out this fact. 

I think you anyway also added a test case for it.

xmxanuel (approved):
looks good.

It makes sense to allow more than auto-increment. :+1:  For stopping multiple signed transactions. 

I think it might be easier to understand if the pattern would be:

```solidity
increaseNonce(uint nonceIncrease)
```
Here the function name contains all information.

Instead of `setNonce`. Here the function name is not full information because you can't set the nonce to any arbitrary value.

CodeSandwich:
`setNonce` is safer than `increaseNonce`, because it's frontrun-proof. Let's say thaty you have 2 signed operations published somewhere. You want to cancel the first one, so you call `increaseNonce(1)`. Now you have no guarantee if the second operation will be cancelled, somebody might execute the first one and you'll cancel the second one. With `setNonce` you are guaranteed to bring the sequence of signatures exactly to the desired state or you get a revert.

xmxanuel:
True, you are right.  This is more safe and allows more precise cancel operations. :+1:

src/Caller.sol
CodeSandwich: That's a good observation, I've been thinking about it for a while and I think that `setNonce` callable by the operators is the right design. It makes `setNonce` batchable and scriptable just like any other operations, which is nice. In case of an emergency (an authorized user going rogue) the original user can always call Caller directly, without a signature which could be blocked, after all only EOAs can generate signed calls. I'm not sure if in the real world this scenario will be common, I can't think of a good reason to have free will (human) operators and using signed calls for delegating execution to anonymous 3rd parties.

PR #288, opened 2023-03-03
Shorten error messages

CodeSandwich:
The error messages now follow the 32-byte recommendation. Also removes a redundant require statement. It cuts the contract size by about 270 bytes allowing higher optimizer runs.


xmxanuel (approved):
looks good.

PR #289, opened 2023-03-07
Add implementation to Managed

CodeSandwich:
This improves discoverability, it's clear what's the current implementation.

### DIFF
--- a/src/Managed.sol
+++ b/src/Managed.sol
@@ -71,6 +71,11 @@ abstract contract Managed is UUPSUpgradeable {
         _managedStorage().isPaused = true;
     }
 
+    /// @notice Returns the current implementation address.
+    function implementation() public view returns (address) {
+        return _getImplementation();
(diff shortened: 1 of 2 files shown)


src/Managed.sol
```diff
@@ -71,6 +71,11 @@ abstract contract Managed is UUPSUpgradeable {
         _managedStorage().isPaused = true;
     }
 
+    /// @notice Returns the current implementation address.
+    function implementation() public view returns (address) {
```
xmxanuel, line 75: Okay make sense. Just, for understanding, we could also just add this function to the proxy itself.  (has the same internal method, needed for the delegateCall).

However, you are saying that in the UUPS pattern, all management functions for the proxy are in the logic contract. 

So, it makes more sense to expose the method in the logic contract as well?

xmxanuel (approved):
looks good

PR #290, opened 2023-03-07
Make admin change two-step

CodeSandwich:
This should make it more difficult to accidentally lose admin. This role is so critical that it probably makes sense to do that, all other serious errors and mistakes can be fixed by the admin.

### DIFF
--- a/src/Managed.sol
+++ b/src/Managed.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.17;
 import {UUPSUpgradeable} from "openzeppelin-contracts/proxy/utils/UUPSUpgradeable.sol";
 import {ERC1967Proxy} from "openzeppelin-contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
+import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 
 using EnumerableSet for EnumerableSet.AddressSet;
 
@@ -17,6 +18,12 @@ using EnumerableSet for EnumerableSet.AddressSet;
 abstract contract Managed is UUPSUpgradeable {
     /// @notice The pointer to the storage slot holding a single `ManagedStorage` structure.
     bytes32 private immutable _managedStorageSlot = _erc1967Slot("eip1967.managed.storage");
+
+    /// @notice Emitted when a new admin of the contract is proposed.
+    /// The proposed admin must call `acceptAdmin` to finalize the change.
+    /// @param currentAdmin The current admin address.
+    /// @param newAdmin The proposed admin address.
+    event NewAdminProposed(address indexed currentAdmin, address indexed newAdmin);
 
     /// @notice Emitted when the pauses role is granted.
     /// @param pauser The address that the pauser role was granted to.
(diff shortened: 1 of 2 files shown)


src/Managed.sol
```diff
@@ -81,10 +89,36 @@ abstract contract Managed is UUPSUpgradeable {
...
+    /// Sets the proposed admin to the zero address.
+    /// Can only be called by the proposed admin.
+    function acceptAdmin() public {
+        require(proposedAdmin() == msg.sender, "Caller not the proposed admin");
+        _managedStorage().proposedAdmin = address(0);
+        _changeAdmin(msg.sender);
+    }
+
+    /// @notice Changes the admin of the contract to address zero.
+    /// It's no longer possible to upgrade the contract afterwards.
```
xmxanuel, line 117:
```suggestion
    /// It's no longer possible to upgrade the contract afterwards if no proposed admin has been set
```
I could still propose another admin before I renounce

src/Managed.sol
```diff
@@ -81,10 +89,36 @@ abstract contract Managed is UUPSUpgradeable {
...
+    function proposedAdmin() public view returns (address) {
+        return _managedStorage().proposedAdmin;
+    }
+
+    /// @notice Proposes a change of the admin of the contract.
+    /// The proposed new admin must call `acceptAdmin` to finalize the change.
+    /// To cancel a proposal propose a different address, e.g. the zero address.
+    /// Can only be called by the current admin.
+    /// @param newAdmin The proposed admin address.
+    function proposeAdmin(address newAdmin) public onlyAdmin {
```
xmxanuel, line 102:
```suggestion
    function proposeNewAdmin(address newAdmin) public onlyAdmin {
```
Maybe?
`proposeAdmin` and `proposedAdmin` are so simliar.

src/Managed.sol
```diff
@@ -81,10 +89,36 @@ abstract contract Managed is UUPSUpgradeable {
...
+    }
+
+    /// @notice Changes the admin of the contract to address zero.
+    /// It's no longer possible to upgrade the contract afterwards.
     /// Can only be called by the current admin.
-    function changeAdmin(address newAdmin) public onlyAdmin {
-        _changeAdmin(newAdmin);
+    function renounceAdmin() public onlyAdmin {
+        emit AdminChanged(admin(), address(0));
+        StorageSlot.getAddressSlot(_ADMIN_SLOT).value = address(0);
```
xmxanuel, line 121: I just looked it up why you are not using the existing function.

```solidity
    function _setAdmin(address newAdmin) private {
        require(newAdmin != address(0), "ERC1967: new admin is the zero address");
        StorageSlot.getAddressSlot(_ADMIN_SLOT).value = newAdmin;
    }

    /**
     * @dev Changes the admin of the proxy.
     *
     * Emits an {AdminChanged} event.
     */
    function _changeAdmin(address newAdmin) internal {
        emit AdminChanged(_getAdmin(), newAdmin);
        _setAdmin(newAdmin);
    }
```

:)

src/Managed.sol
CodeSandwich: Good idea, applied.

CodeSandwich:
I think that we don't need `denyProposedAdmin`. It's basically the same as `proposeNewAdmin(address(0))`, the added explicitness is nice, but it's probably not worth the ABI surface, plus we're really stretching the contract bytecode size and each public function takes space.

PR #296, opened 2023-03-10
Add minting NFTs with salt

CodeSandwich:


src/NFTDriver.sol
```diff
@@ -30,8 +30,16 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Managed {
...
-    bytes32 private immutable _mintedTokensSlot = _erc1967Slot("eip1967.nftDriver.storage");
+    /// @notice The ERC-1967 storage slot holding a single `NFTDriverStorage` structure.
+    bytes32 private immutable _nftDriverStorageSlot = _erc1967Slot("eip1967.nftDriver.storage");
+
+    struct NFTDriverStorage {
+        /// @notice The number of tokens minted without salt.
+        uint64 mintedTokens;
+        /// @notice The salts already used for minting tokens.
+        /// The keys are the minter address and the salt, the value is true if the salt is used.
+        mapping(address => mapping(uint64 => bool)) isSaltUsed;
```
xmxanuel, line 41: Why is this needed? We could just use the ERC721 `_exists` function and check if a token exists.

src/NFTDriver.sol
```diff
@@ -55,7 +63,30 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Managed {
...
+    /// @param minter The minter of the token.
+    /// @param salt The salt used for minting the token.
+    /// @return tokenId The token ID.
+    function calcTokenIdWithSalt(address minter, uint64 salt)
+        public
+        view
+        returns (uint256 tokenId)
+    {
+        tokenId = driverId;
+        tokenId = (tokenId << 160) | uint160(minter);
```
xmxanuel, line 79: I think it would be nice to use constants for the offsets

src/NFTDriver.sol
```diff
@@ -55,7 +63,30 @@ contract NFTDriver is ERC721Burnable, ERC2771Context, Managed {
...
+        tokenId = (tokenId << 64) | salt;
+    }
+
+    /// @notice Checks if the salt has already been used for minting a token.
+    /// Each minter can use each salt only once, to mint a single token.
+    /// @param minter The minter of the token.
+    /// @param salt The salt used for minting the token.
+    /// @return isUsed True if the salt has been used, false otherwise.
+    function isSaltUsed(address minter, uint64 salt) public view returns (bool isUsed) {
```
xmxanuel, line 88: We can just make `exists` public

xmxanuel (commented):
In general, it looks good. I think we could simplify it a bit with the exists function.

src/NFTDriver.sol
xmxanuel: We could also keep this function but build it on top of `exists`.

src/NFTDriver.sol
xmxanuel: We should think about the behaviour we would like to see together with `burn`.

xmxanuel: What should happen after a token burn. Should it be possible to mint it again? The current design wouldn't allow that.

src/NFTDriver.sol
CodeSandwich: I've looked into this and `exists` is reset after burning a token. If we want burning to be permanent, we can't use `exists`.

src/NFTDriver.sol
CodeSandwich: I've tried different approaches and this one seemed the cleanest and most concise. Could you propose changes with the alternative design?

src/NFTDriver.sol
xmxanuel: Yes, I just realized otherwise it would result in unexpected behavior. 
For example, the `minter` could have transferred the token to Alice. Alice wants to burn the token some funds regardless that some funds are still waiting for here.

In that case,  it would be not okay if the `minter` can just `mint` the same `tokenId` again and access the funds.

I see two options:
a) the additional mapping for the `salt` which is not reset in the `burn` case. which effectively prevents the re-mint
b) remove the ability to `burn` tokens at all and safe some gas

src/NFTDriver.sol
xmxanuel:
```solidity
uint internal constant ADDRESS_OFFSET = 160;
```
```solidity
 tokenId = (tokenId << ADDRESS_OFFSET) | uint160(minter);
```

src/NFTDriver.sol
xmxanuel: Yes, you are right. I came to the conlusion if we `burn` it should be not possible to re-mint

src/NFTDriver.sol
CodeSandwich: This PR goes with the `a)` option. IMO `b)` is a bit heavy handed, official burning is a great way to officially signal abandonment and freezing of a user ID.

src/NFTDriver.sol
CodeSandwich: Is it cleaner or more bug-resistant? To me it seems noisy and in non-local, we get an extra constant to look at and you need to find the value of `ADDRESS_OFFSET` to see if the code makes sense. The expression `(tokenId << 160) | uint160(minter);` is self-contained and obviously correct, shift by 160 bits and set 160 bits. If we needed to repeat this logic in multiple places, I'd agree, a constant would be a good idea, but here it's just this one place.

xmxanuel: In my opinion, it makes the code easier to understand.  Currently, you leave it as an exercise to the reader.
The reader needs to realize, ah 160 bits are the offset for an address.

It is easier in my opinion if I can read the interpretation of a number.  

This is my reasoning behind it.

CodeSandwich: If it's unclear, maybe we should comment the shifting operations? The solution with the constant only adds one piece of information, that 160 is the "address offset", but it spreads it across the entire contract, a comment would be available locally.

xmxanuel: I think we should add a comment. We had C4 auditors who struggled with bit manipulation and shifting. A lot of frontend dev's might be also not so familiar with these low-level concepts.

CodeSandwich: Ok, I'll add docs in the next PR, there are a few places with bit shifting, so it makes sense to comment them all.

xmxanuel (approved):
looks good

PR #297, opened 2023-03-13
Add deployer contract

CodeSandwich:
Makes deployment a single transaction. It requires less than 18M gas, which is fine even on chains with lower gas limits like Optimism (checked on its testnet, it works). Stores the deployment details on chain and updates the JSON. Makes deployment easily repeatable on different chains. Reduces the number of deployment options, now the entire set of contracts is always deployed. Adds tests for the deployment.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -10,181 +10,117 @@ print_title() {
     echo
 }
 
-create() {
-    print_title "Creating $1"
-    DEPLOYED_ADDR=$( \
-        forge create $VERIFY $WALLET_ARGS "$2" --constructor-args "${@:3}" \
-        | tee /dev/tty | grep '^Deployed to: ' | cut -d " " -f 3)
-}
-
-send() {
-    print_title "$1"
-    cast send $WALLET_ARGS "$2" "$3" "${@:4}"
-}
-
-assertIsDriver() {
-    local DRIVER_ID=$(cast call "$2" 'driverId()(uint32)')
-    local DRIVER_ID_ADDR=$(cast call "$DRIPS_HUB" 'driverAddress(uint32)(address)' "$DRIVER_ID")
-    if [ $(cast --to-checksum-address "$2") != "$DRIVER_ID_ADDR" ]; then
-        echo
-        echo "$1 not registered as a driver in DripsHub"
-        echo "DripsHub address: $DRIPS_HUB"
-        echo "$1 ID: $DRIVER_ID"
-        echo "$1 address: $DRIVER"
-        echo "Driver address registered under ID $DRIVER_ID: $DRIVER_ID_ADDR"
-        exit 2
+# Verify parameters
+verify_parameter() {
+    if [ -z "${!1}" ]; then
+        echo Error: "'$1'" variable not set, see README.md
+        exit 1
     fi
 }
+verify_parameter ETH_RPC_URL
+verify_parameter WALLET_ARGS
 
 # Set up the defaults
-NETWORK=$(cast chain)
-DEPLOYMENT_JSON=${DEPLOYMENT_JSON:-./deployment_$NETWORK.json}
-DEPLOYER=$(cast wallet address $WALLET_ARGS | cut -d " " -f 2)
-GOVERNANCE=${GOVERNANCE:-$DEPLOYER}
-DRIPS_HUB_ADMIN=$(cast --to-checksum-address "${DRIPS_HUB_ADMIN:-$GOVERNANCE}")
-ADDRESS_DRIVER_ADMIN=$(cast --to-checksum-address "${ADDRESS_DRIVER_ADMIN:-$GOVERNANCE}")
-NFT_DRIVER_ADMIN=$(cast --to-checksum-address "${NFT_DRIVER_ADMIN:-$GOVERNANCE}")
-SPLITS_DRIVER_ADMIN=$(cast --to-checksum-address "${SPLITS_DRIVER_ADMIN:-$GOVERNANCE}")
-CYCLE_SECS=${CYCLE_SECS:-$(( 7 * 24 * 60 * 60 ))} # 1 week
-if [ -n "$ETHERSCAN_API_KEY" ]; then
-    VERIFY="--verify"
-else
-    VERIFY=""
-fi
+CHAIN=$(cast chain)
+DEPLOYMENT_JSON=${DEPLOYMENT_JSON:-./deployment_$CHAIN.json}
+WALLET=$(cast wallet address $WALLET_ARGS | cut -d " " -f 2)
+WALLET_NONCE=$(cast nonce "$WALLET")
+ADMIN=${ADMIN:-$WALLET}
+DRIPS_HUB_ADMIN=$(cast --to-checksum-address "${DRIPS_HUB_ADMIN:-$ADMIN}")
+ADDRESS_DRIVER_ADMIN=$(cast --to-checksum-address "${ADDRESS_DRIVER_ADMIN:-$ADMIN}")
+NFT_DRIVER_ADMIN=$(cast --to-checksum-address "${NFT_DRIVER_ADMIN:-$ADMIN}")
+SPLITS_DRIVER_ADMIN=$(cast --to-checksum-address "${SPLITS_DRIVER_ADMIN:-$ADMIN}")
+DRIPS_HUB_CYCLE_SECS=${DRIPS_HUB_CYCLE_SECS:-$(( 7 * 24 * 60 * 60 ))} # 1 week
 
 # Print the configuration
-print_title "Deployment Config"
-echo "Network:                      $NETWORK"
-echo "Deployer address:             $DEPLOYER"
+print_title Deployment configuration
+echo "Chain:                        $CHAIN"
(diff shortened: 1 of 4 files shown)


src/Deployer.sol
```diff
@@ -0,0 +1,130 @@
...
+    function deployCaller() internal {
+        caller = new Caller();
+        callerArgs = abi.encode();
+    }
+
+    /// @dev Requires DripsHub and Caller to be deployed
+    function deployAddressDriver(address addressDriverAdmin_) internal {
+        // Deploy logic
+        address forwarder = address(caller);
+        uint32 driverId = dripsHub.nextDriverId();
```
xmxanuel, line 83:
```suggestion
        uint32 driverId = dripsHub.nextDriverId();
        require(driverId == 0, "incorrect driverId for AddressDriver");
```
It is not needed but I think it is nice to verify some rules

src/Deployer.sol
```diff
@@ -0,0 +1,130 @@
...
+    address public immutableSplitsDriverAdmin;
+    uint32 public immutableSplitsDriverId;
+
+    constructor(
+        uint32 dripsHubCycleSecs_,
+        address dripsHubAdmin_,
+        address addressDriverAdmin_,
+        address nftDriverAdmin_,
+        address immutableSplitsDriverAdmin_
+    ) {
```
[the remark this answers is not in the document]
xmxanuel: Why three different admins? I think for now we could simplify it. 
We only need one.

scripts/deploy.sh
```diff
@@ -10,74 +10,34 @@ print_title() {
...
+DRIPS_HUB_CYCLE_SECS=${DRIPS_HUB_CYCLE_SECS:-$(( 7 * 24 * 60 * 60 ))} # 1 week
 
 # Print the configuration
-print_title "Deployment Config"
-echo "Network:                      $NETWORK"
-echo "Deployer address:             $DEPLOYER"
+print_title Deployment configuration
+echo "Chain:                        $CHAIN"
+echo "Wallet:                       $WALLET"
+echo "Wallet nonce:                 $WALLET_NONCE"
```
xmxanuel, line 29: :+1:

scripts/deploy.sh
```diff
@@ -87,104 +47,69 @@ then
...
-assertIsDriver "ImmutableSplitsDriver" "$SPLITS_DRIVER"
+DEPLOYER=$( \
+    forge create $VERIFY $WALLET_ARGS "src/Deployer.sol:Deployer" --constructor-args \
+        "$DRIPS_HUB_CYCLE_SECS" \
+        "$DRIPS_HUB_ADMIN" \
+        "$ADDRESS_DRIVER_ADMIN" \
+        "$NFT_DRIVER_ADMIN" \
+        "$SPLITS_DRIVER_ADMIN" \
+    | tee /dev/tty | grep '^Deployed to: ' | cut -d " " -f 3)
+deployment_detail() {
```
xmxanuel, line 65: I like your helper bash function. :sun_with_face: 
A parameter for defining the return type for the contract call.

We avoid the helper variable so it is good

scripts/deploy.sh
```diff
@@ -87,104 +47,69 @@ then
...
+    "AddressDriver ID":             "$(deployment_detail addressDriverId uint32)",
+    "NFTDriver":                    "$(deployment_detail nftDriver address)",
+    "NFTDriver logic":              "$(deployment_detail nftDriverLogic address)",
+    "NFTDriver admin":              "$(deployment_detail nftDriverAdmin address)",
+    "NFTDriver ID":                 "$(deployment_detail nftDriverId uint32)",
+    "ImmutableSplitsDriver":        "$(deployment_detail immutableSplitsDriver address)",
+    "ImmutableSplitsDriver logic":  "$(deployment_detail immutableSplitsDriverLogic address)",
+    "ImmutableSplitsDriver admin":  "$(deployment_detail immutableSplitsDriverAdmin address)",
+    "ImmutableSplitsDriver ID":     "$(deployment_detail immutableSplitsDriverId uint32)",
     "Commit hash":                  "$(git rev-parse HEAD)"
```
xmxanuel, line 113: Since we trying to make this perfect here. We could also add a deployment timestamp. :)

xmxanuel (commented):
In general, it looks very good. I added a few questions and ideas.

src/Deployer.sol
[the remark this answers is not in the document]
CodeSandwich: > We could design is already in a more create2 readiness.

But why? We don't need CREATE2, it doesn't provide us anything while making deployment addresses **more** volatile because dependent on the deployed bytecode. If deployment is callable only once, we know upfront what's the nonce and we don't care about the bytecode.

> deploy method which is only callable once

We can't do that, because it'd require Deployer to contain a copy of bytecode of all the deployed contracts. It'd more or less double the deployment cost. Most importantly it'd be tricky to fit so much data in a single contract bytecode, it'd need to be somehow split.

> Why three different admins?

I didn't want to drop the ability to make admins different, e.g. owned by different addresses or made non-upgradeable. On one hand you're right, we probably are going to use one address in practice, but on the other it doesn't take much to support varied ownership. WDYT?

src/Deployer.sol
```diff
@@ -0,0 +1,130 @@
...
+
+contract Deployer {
+    // slither-disable-next-line immutable-states
+    address public creator;
+
+    DripsHub public dripsHub;
+    bytes public dripsHubArgs;
+    uint32 public dripsHubCycleSecs;
+    DripsHub public dripsHubLogic;
+    bytes public dripsHubLogicArgs;
```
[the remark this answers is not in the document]
CodeSandwich: The deployments are internal transactions, so it's not easy to get their arguments. Foundry only fetches the whole transaction arguments, I think that even the raw Etherscan API doesn't expose data like this. This would be fine, but we need the arguments for contracts verification on Etherscan.

src/Deployer.sol
```diff
@@ -0,0 +1,130 @@
...
+    ) {
+        creator = msg.sender;
+        deployDripsHub(dripsHubCycleSecs_, dripsHubAdmin_);
+        deployCaller();
+        deployAddressDriver(addressDriverAdmin_);
+        deployNFTDriver(nftDriverAdmin_);
+        deployImmutableSplitsDriver(immutableSplitsDriverAdmin_);
+    }
+
+    function deployDripsHub(uint32 dripsHubCycleSecs_, address dripsHubAdmin_) internal {
```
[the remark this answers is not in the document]
CodeSandwich: Good idea, fixed

src/Deployer.sol
```diff
@@ -0,0 +1,130 @@
...
+        dripsHubAdmin = dripsHubAdmin_;
+        // slither-disable-next-line reentrancy-benign
+        ManagedProxy proxy = new ManagedProxy(dripsHubLogic, dripsHubAdmin);
+        dripsHub = DripsHub(address(proxy));
+        dripsHubArgs = abi.encode(dripsHubLogic, dripsHubAdmin);
+    }
+
+    function deployCaller() internal {
+        caller = new Caller();
+        callerArgs = abi.encode();
```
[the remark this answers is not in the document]
CodeSandwich: For consistency :sweat_smile:

src/Deployer.sol
CodeSandwich: We do that in the tests, it makes sure that any changes in the source lead to the right IDs. I don't think that we need that in the code itself, to me it seems noisy and redundant, we're running it on exactly the same smart contracts as tests.

xmxanuel: Yeah, I saw that. I mean it is not required.

src/Deployer.sol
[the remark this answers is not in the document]
xmxanuel: We could reconstruct it in the bash script (since we know the params) but it might be cleaner this way.

src/Deployer.sol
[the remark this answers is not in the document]
xmxanuel: > We can't do that, because it'd require Deployer to contain a copy of bytecode of all the deployed contracts. It'd more or less double the deployment cost. Most importantly it'd be tricky to fit so much data in a single contract bytecode, it'd need to be somehow split.

Ah I see. I was not considering this point. Then it makes sense to trigger it in the constructor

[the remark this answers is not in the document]
xmxanuel: I think we can also keep the three admins instead of changing it again.

xmxanuel (approved):
lgtm

scripts/deploy.sh
CodeSandwich: It's doable, but do we need it? When the JSON is read, the contracts are already deployed. If you're just curious, all the data is on-chain. It's not critical information from the trust and security perspective either.

xmxanuel: Yes, of course, you can fetch it from the transactions. If we end up multiple of this json files (different deployment times) it might be nice to have it.

CodeSandwich: That's a good point. It doesn't require blockchain-perfect transaction timestamp, so it can be much simpler. I've added it using a regular time from the local machine, it's very easy and for comparing different JSONs it should be enough.

PR #298, opened 2023-03-14
Improve comments

CodeSandwich:
Braks bitwise operations into simple, commented steps. Adds names to mapping keys. Adds a few missing doc comments.


xmxanuel (approved):
nice improvement

PR #299, opened 2023-04-24
Add RepoDriver

CodeSandwich:
The production logic is ready to be merged, there are no tests yet, working on them.

Tested out on Goerli (0x45Aa599e4EB38300025eAB83908652bcC61601D4) with GitHub and GitLab, both with the driver covering the fees and the user with `transferAndCall`.

### DIFF
--- a/src/NFTDriver.sol
+++ b/src/NFTDriver.sol
@@ -11,7 +11,6 @@ import {
 } from "./DripsHub.sol";
 import {Managed} from "./Managed.sol";
 import {Context, ERC2771Context} from "openzeppelin-contracts/metatx/ERC2771Context.sol";
-import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
 import {
     ERC721,
     ERC721Burnable,
--- a/src/RepoDriver.sol
+++ b/src/RepoDriver.sol
@@ -1,0 +1,536 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.19;
+
+import {
+    DripsHub,
+    DripsReceiver,
+    IERC20,
+    SafeERC20,
+    SplitsReceiver,
+    UserMetadata
+} from "./DripsHub.sol";
+import {Managed} from "./Managed.sol";
+import {ERC677ReceiverInterface} from "chainlink/interfaces/ERC677ReceiverInterface.sol";
+import {LinkTokenInterface} from "chainlink/interfaces/LinkTokenInterface.sol";
+import {OperatorInterface} from "chainlink/interfaces/OperatorInterface.sol";
+import {BufferChainlink, CBORChainlink} from "chainlink/Chainlink.sol";
+import {Context, ERC2771Context} from "openzeppelin-contracts/metatx/ERC2771Context.sol";
+
+/// @notice The supported forges where repositories are stored.
+enum Forge {
+    GitHub,
+    GitLab
+}
+
+/// @notice A DripsHub driver implementing repository-based user identification.
+/// Each repository stored in one of the supported forges has a deterministic user ID assigned.
+/// By default the repositories have no owner and their users can't be controlled by anybody,
+/// use `requestUpdateOwner` to update the owner.
+contract RepoDriver is ERC677ReceiverInterface, ERC2771Context, Managed {
+    using SafeERC20 for IERC20;
+    using CBORChainlink for BufferChainlink.buffer;
+
+    uint256 internal constant CHAIN_ID_ETHEREUM = 1;
+    string internal constant CHAIN_NAME_ETHEREUM = "ethereum";
+    address internal constant LINK_TOKEN_ETHEREUM = 0x514910771AF9Ca656af840dff83E8264EcF986CA;
+
+    uint256 internal constant CHAIN_ID_GOERLI = 5;
+    string internal constant CHAIN_NAME_GOERLI = "goerli";
+    address internal constant LINK_TOKEN_GOERLI = 0x326C977E6efc84E512bB9C30f76E30c160eD06FB;
+
+    uint256 internal constant CHAIN_ID_SEPOLIA = 11155111;
+    string internal constant CHAIN_NAME_SEPOLIA = "sepolia";
+    address internal constant LINK_TOKEN_SEPOLIA = 0x779877A7B0D9E8603169DdbD7836e478b4624789;
+
+    /// @notice The DripsHub address used by this driver.
+    DripsHub public immutable dripsHub;
+    /// @notice The driver ID which this driver uses when calling DripsHub.
+    uint32 public immutable driverId;
+    /// @notice The Link token used for paying the operators.
+    LinkTokenInterface public immutable linkToken;
+
+    /// @notice The ERC-1967 storage slot holding a single `RepoDriverStorage` structure.
+    bytes32 private immutable _repoDriverStorageSlot = _erc1967Slot("eip1967.repoDriver.storage");
+    /// @notice The ERC-1967 storage slot holding a single `RepoDriverAnyApiStorage` structure.
+    bytes32 private immutable _repoDriverAnyApiStorageSlot =
+        _erc1967Slot("eip1967.repoDriver.anyApi.storage");
+
+    /// @notice Emitted when the AnyApi operator configuration is updated.
+    /// @param operator The new address of the AnyApi operator.
+    /// @param jobId The new AnyApi job ID used for requesting user owner updates.
+    /// @param defaultFee The new fee in Link for each user owner
+    /// update request when the driver is covering the cost.
+    event AnyApiOperatorUpdated(
+        OperatorInterface indexed operator, bytes32 indexed jobId, uint96 defaultFee
+    );
+
+    /// @notice Emitted when the user ownership update is requested.
+    /// @param userId The ID of the user.
+    /// @param forge The forge where the repository is stored.
+    /// @param name The name of the repository.
+    event OwnerUpdateRequested(uint256 indexed userId, Forge forge, bytes name);
+
+    /// @notice Emitted when the user ownership is updated.
+    /// @param userId The ID of the user.
+    /// @param owner The new owner of the repository.
+    event OwnerUpdated(uint256 indexed userId, address owner);
+
+    struct RepoDriverStorage {
+        /// @notice The owners of the users.
+        mapping(uint256 userId => address) userOwners;
+    }
+
+    struct RepoDriverAnyApiStorage {
+        /// @notice The requested user owner updates.
+        mapping(bytes32 requestId => uint256 userId) requestedUpdates;
+        /// @notice The new address of the AnyApi operator.
+        OperatorInterface operator;
+        /// @notice The fee in Link for each user owner
+        /// update request when the driver is covering the cost.
+        uint96 defaultFee;
+        /// @notice The AnyApi job ID used for requesting user owner updates.
+        bytes32 jobId;
+        /// @notice The AnyApi requests counter used as a nonce when calculating the request ID.
+        uint256 nonce;
+    }
+
+    /// @param _dripsHub The drips hub to use.
+    /// @param forwarder The ERC-2771 forwarder to trust. May be the zero address.
+    /// @param _driverId The driver ID to use when calling DripsHub.
+    constructor(DripsHub _dripsHub, address forwarder, uint32 _driverId)
+        ERC2771Context(forwarder)
+    {
+        dripsHub = _dripsHub;
+        driverId = _driverId;
+        // slither-disable-next-line uninitialized-local
+        address _linkToken;
+        if (block.chainid == CHAIN_ID_ETHEREUM) _linkToken = LINK_TOKEN_ETHEREUM;
+        else if (block.chainid == CHAIN_ID_GOERLI) _linkToken = LINK_TOKEN_GOERLI;
+        else if (block.chainid == CHAIN_ID_SEPOLIA) _linkToken = LINK_TOKEN_SEPOLIA;
+        else revert("Unsupported chain");
+        linkToken = LinkTokenInterface(_linkToken);
+    }
+
+    modifier onlyOwner(uint256 userId) {
+        require(_msgSender() == ownerOf(userId), "Caller is not the user owner");
+        _;
+    }
+
+    /// @notice Calculates the user ID.
+    /// Every user ID is a 256-bit integer constructed by concatenating:
+    /// `driverId (32 bits) | forgeId (8 bits) | nameEncoded (216 bits)`.
+    /// When `forge` is GitHub and `name` is at most 27 bytes long,
+    /// `forgeId` is 0 and `nameEncoded` is `name` right-padded with zeros
+    /// When `forge` is GitHub and `name` is longer than 27 bytes,
+    /// `forgeId` is 1 and `nameEncoded` is the lower 27 bytes of the hash of `name`.
+    /// When `forge` is GitLab and `name` is at most 27 bytes long,
+    /// `forgeId` is 2 and `nameEncoded` is `name` right-padded with zeros
+    /// When `forge` is GitLab and `name` is longer than 27 bytes,
+    /// `forgeId` is 3 and `nameEncoded` is the lower 27 bytes of the hash of `name`.
+    /// @param forge The forge where the repository is stored.
+    /// @param name The name of the repository.
+    /// For GitHub and GitLab it must follow the `user_name/repository_name` structure
+    /// and it must be formatted identically as in the repository's URL,
+    /// including the case of each letter and special characters being removed.
+    /// @return userId The user ID.
+    function calcUserId(Forge forge, bytes memory name) public view returns (uint256 userId) {
+        uint8 forgeId;
+        uint216 nameEncoded;
+        if (forge == Forge.GitHub) {
+            if (name.length <= 27) {
+                forgeId = 0;
+                nameEncoded = uint216(bytes27(name));
+            } else {
+                forgeId = 1;
+                // `nameEncoded` is the lower 27 bytes of the hash
+                nameEncoded = uint216(uint256(keccak256(name)));
+            }
+        } else {
+            if (name.length <= 27) {
+                forgeId = 2;
+                nameEncoded = uint216(bytes27(name));
+            } else {
+                forgeId = 3;
+                // `nameEncoded` is the lower 27 bytes of the hash
+                nameEncoded = uint216(uint256(keccak256(name)));
+            }
+        }
+        // By assignment we get `userId` value:
+        // `zeros (224 bits) | driverId (32 bits)`
+        userId = driverId;
+        // By bit shifting we get `userId` value:
+        // `zeros (216 bits) | driverId (32 bits) | zeros (8 bits)`
+        // By bit masking we get `userId` value:
+        // `zeros (216 bits) | driverId (32 bits) | forgeId (8 bits)`
+        userId = (userId << 8) | forgeId;
+        // By bit shifting we get `userId` value:
+        // `driverId (32 bits) | forgeId (8 bits) | zeros (216 bits)`
+        // By bit masking we get `userId` value:
+        // `driverId (32 bits) | forgeId (8 bits) | nameEncoded (216 bits)`
+        userId = (userId << 216) | nameEncoded;
+    }
+
+    /// @notice Updates the AnyApi operator configuration. Callable only by the admin.
+    /// @param operator The new address of the AnyApi operator.
+    /// @param jobId The new AnyApi job ID used for requesting user owner updates.
+    /// @param defaultFee The new fee in Link for each user owner
+    /// update request when the driver is covering the cost.
+    function updateAnyApiOperator(OperatorInterface operator, bytes32 jobId, uint96 defaultFee)
+        public
+        whenNotPaused
+        onlyAdmin
+    {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        storageRef.operator = operator;
+        storageRef.jobId = jobId;
+        storageRef.defaultFee = defaultFee;
+        emit AnyApiOperatorUpdated(operator, jobId, defaultFee);
+    }
+
+    /// @notice Gets the current AnyApi operator configuration.
+    /// @return operator The address of the AnyApi operator.
+    /// @return jobId The AnyApi job ID used for requesting user owner updates.
+    /// @return defaultFee The fee in Link for each user owner
+    /// update request when the driver is covering the cost.
+    function anyApiOperator()
+        public
+        view
+        returns (OperatorInterface operator, bytes32 jobId, uint96 defaultFee)
+    {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        operator = storageRef.operator;
+        jobId = storageRef.jobId;
+        defaultFee = storageRef.defaultFee;
+    }
+
+    /// @notice Gets the user owner.
+    /// @param userId The ID of the user.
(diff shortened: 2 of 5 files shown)


src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    /// @param name The name of the repository.
+    /// @return url The built URL.
+    function _requestUrl(Forge forge, bytes memory name)
+        internal
+        pure
+        returns (string memory url)
+    {
+        if (forge == Forge.GitHub) {
+            return string.concat(
+                "https://raw.githubusercontent.com/", string(name), "/HEAD/FUNDING.json"
```
xmxanuel, line 292: Ah nice, I didn't know that the `HEAD` will just always result in the default branch. This simplifies it a lot.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+
+    /// @notice Builds the JSON path inside `FUNDING.json` where the repository owner is stored.
+    /// @return path The comma-separated JSON path.
+    function _requestPath() internal view returns (string memory path) {
+        // slither-disable-next-line uninitialized-local
+        string memory chainName;
+        if (block.chainid == CHAIN_ID_ETHEREUM) chainName = CHAIN_NAME_ETHEREUM;
+        else if (block.chainid == CHAIN_ID_GOERLI) chainName = CHAIN_NAME_GOERLI;
+        else revert("Unsupported chain");
+        return string.concat("drips,", chainName, ",ownedBy");
```
xmxanuel, line 309: I wouldn't call it `ownedBy`. Technical this is correct, that the `repoId` is controlled by this address.
The github repo on the other hand is not 'owned' by this address.
However, from the project perspective it is more `projectAddress` or something.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    /// @param ownerRaw The new owner of the repository. Must be a 20 bytes long address.
+    function updateRepoOwnerByAnyApi(bytes32 requestId, bytes calldata ownerRaw) public {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        require(msg.sender == address(storageRef.operator), "Callable only by the operator");
+        uint224 repoId = storageRef.requestedUpdates[requestId];
+        require(repoId != 0, "Unknown request ID");
+        delete storageRef.requestedUpdates[requestId];
+        require(ownerRaw.length == 20, "Invalid owner length");
+        address owner = address(bytes20(ownerRaw));
+        _repoDriverStorage().repoOwners[repoId] = owner;
```
xmxanuel, line 325: One idea. Together, with the `owner` we also store a `lastUpdated` and add some delay period. 

An update is only possible once every 30 minutes or something.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    /// @notice Updates the repository owner. Callable only by the AnyApi operator.
+    /// @param requestId The ID of the AnyApi request.
+    /// Must be the same as the request ID generated when requesting an owner update,
+    /// this function will update the repository that was requested back then.
+    /// @param ownerRaw The new owner of the repository. Must be a 20 bytes long address.
+    function updateRepoOwnerByAnyApi(bytes32 requestId, bytes calldata ownerRaw) public {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        require(msg.sender == address(storageRef.operator), "Callable only by the operator");
```
xmxanuel, line 319: Here, we could also use the trusted forwarder but I guess chainlink operators won't need it.  :)

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    /// @param jobId The new AnyApi job ID used for requesting repository owner updates.
+    /// @param defaultFee The new fee in Link for each repository owner
+    /// update request when the driver is covering the cost.
+    function updateAnyApiOperator(OperatorInterface operator, bytes32 jobId, uint96 defaultFee)
+        public
+        whenNotPaused
+        onlyAdmin
+    {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        storageRef.operator = operator;
```
xmxanuel, line 160: Do you think it will always be just one operator? We could also use a `mapping` for operators.

xmxanuel (commented):
nice work. Looks already really good.

I added some comments and ideas.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+        ERC2771Context(forwarder)
+    {
+        dripsHub = _dripsHub;
+        driverId = _driverId;
+        // slither-disable-next-line uninitialized-local
+        address _linkToken;
+        if (block.chainid == CHAIN_ID_ETHEREUM) _linkToken = LINK_TOKEN_ETHEREUM;
+        else if (block.chainid == CHAIN_ID_GOERLI) _linkToken = LINK_TOKEN_GOERLI;
+        else revert("Unsupported chain");
+        linkToken = LinkTokenInterface(_linkToken);
```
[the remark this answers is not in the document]
CodeSandwich: That's a really cool idea :thinking: The only downside is that then the deployer would need to be extra careful to deploy the right version to the right network.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    /// `forge (8 bits) | uint216(keccak256(name)) (216 bits)`.
+    /// @param forge The forge where the repository is stored.
+    /// @param name The name of the repository.
+    /// @return repoId The ID of the repository.
+    function calcRepoId(Forge forge, bytes memory name) public pure returns (uint224 repoId) {
+        // By assignment we get `repoId` value:
+        // `zeros (216 bits) | forge (8 bits)`
+        repoId = uint8(forge);
+        // `nameHash` is the lower 27 bytes of the hash
+        uint216 nameHash = uint216(uint256(keccak256(name)));
```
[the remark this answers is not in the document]
CodeSandwich: The Ethereum address is a 160-bit hash and if one can find a colliding preimage for a given address, it's much worse than here, 216 should be more than enough. This design allows having different repoId generation schemes for different forges. E.g. for GitHub it's a hash of the name, but for Heartwood it may be just the last 216 bits of the repo ID. On top of that you're right, it adds more information readily available in the repoId without making the design insecure or adding complexity, the only difference is whether you append the forge before or after hashing.

src/RepoDriver.sol
```diff
@@ -0,0 +1,508 @@
...
+    /// @notice Requests the AnyApi operator to update the repository owner.
+    /// See `requestUpdateRepoOwner` for more details.
+    /// @param forge The forge where the repository is stored.
+    /// @param name The name of the repository.
+    /// @param fee The fee in Link to pay for the request.
+    function _requestUpdateRepoOwner(Forge forge, bytes memory name, uint256 fee) internal {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        address operator = address(storageRef.operator);
+        require(operator != address(0), "Operator address not set");
+        uint256 nonce = storageRef.nonce++;
```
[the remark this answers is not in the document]
CodeSandwich: We must allow multiple pending requests. If a request is invalid, e.g. because too little Link has been paid for it, it's not explicitly cancelled, it's silently ignored by Chainlink. I could front run your verification request with the same request, but with too little Link attached. That would block your good request, probably forever.

Chainlink has no order guarantee. The good thing is that if there are multiple requests for the same repo, the last executed one will also be based on the freshest state of the repo, which is good, just wasteful, I don't think that it's exploitable. The only attack would be if the operator cached an old result and published it after you updated the owner reverting your change. But if the operator is malicious, they can do anything they want anyway.

src/RepoDriver.sol
CodeSandwich: Yes, it does, I didn't know this either and was so happy to discover this :raised_hands:

src/RepoDriver.sol
CodeSandwich: You're right, it could be understood as the GitHub or GitLab repo ownership. So maybe `controlledBy`? OTOH it lives under `drips->ethereum->ownedBy`, so it's clear that it's the ownership in the context of Drips. `projectAddress` seems confusing, many users will just paste their wallet address there or the team's mulstisig, which is not a project.

src/RepoDriver.sol
CodeSandwich: What would it protect from? It certainly would be annoying, people will be making mistakes and many of them will be requesting multiple updates in the span of minutes when they're doing the initial configuration.

src/RepoDriver.sol
CodeSandwich: No, we shouldn't do that. Especially since there may be a possibility to trick the Chainlink operator to call Caller as a part of the AnyApi response and grant the attacker full rights to act as the operator.

src/RepoDriver.sol
CodeSandwich: Yes, I think so, why would there be more and how would they be chosen?

src/RepoDriver.sol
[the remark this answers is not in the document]
xmxanuel: It is a nice pattern, if each part of the code gets used. 
Here it is either `if` or `else` but never both.

src/RepoDriver.sol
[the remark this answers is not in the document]
xmxanuel: There are some discussions around it: 
https://ethereum-magicians.org/t/increasing-address-size-from-20-to-32-bytes/5485/6

src/RepoDriver.sol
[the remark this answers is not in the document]
xmxanuel: Ah true. Then we would have to implement some `cancel` of pending transaction which still would delay the entire process.

src/RepoDriver.sol
xmxanuel: Yeah, no strong opinion.

src/RepoDriver.sol
xmxanuel: In case one is down or something. Let's keep it for now with just one.

src/RepoDriver.sol
xmxanuel: Yeah that's a good point.

xmxanuel (approved):
looks good to me

src/RepoDriver.sol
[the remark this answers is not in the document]
CodeSandwich: If anybody finds a way to generate a preimage of a 160-bit keccak, they will generate a private key for `DripsHub`'s address and just steal all the tokens held by it anyway :shrug: The 216-bit repo identity will be a hollow fortress burned from the inside and abandoned. Making it a 224-bit or even a 256-bit won't save it.

src/RepoDriver.sol
CodeSandwich: I'd stick with `ownedBy`, it's a common concept in crypto and `controlledBy` doesn't clear any confusion about the context.

src/RepoDriver.sol
CodeSandwich: It seems that the uptime of the operators is pretty good. If ours is down, we won't find out until valid requests stop being responded to, the current logic of the operator contract doesn't support the state of being down. We'd need to have a list of supported operators with the users being able to pick one for the given request, they would have to detect something being wrong and retry with the different operator. It would work, but I don't think that it's worth developing right now, AnyApi isn't even the ultimate solution.

CodeSandwich:
Pushed a lot of changes, added tests, cleaned up the API and added support for plain text short repo names in user IDs.

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+    /// For GitHub and GitLab it must follow the `user_name/repository_name` structure
+    /// and it must be formatted identically as in the repository's URL,
+    /// including the case of each letter and special characters being removed.
+    /// @return userId The user ID.
+    function calcUserId(Forge forge, bytes memory name) public view returns (uint256 userId) {
+        uint8 forgeId;
+        uint216 nameEncoded;
+        if (forge == Forge.GitHub) {
+            if (name.length <= 27) {
+                forgeId = 0;
```
xmxanuel, line 136: The enum seems a bit over-engineered, if we still need a `uint8` helper variable here.
```solidity
function calcUserId(uint8 forge, bytes memory name)
```

Would avoid it. We could introduce constants `FORGE_GITHUB`.

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+    /// to determine which user's ownership update is requested.
+    function onTokenTransfer(address, /* sender */ uint256 amount, bytes calldata data)
+        public
+        whenNotPaused
+    {
+        require(msg.sender == address(linkToken), "Callable only by the Link token");
+        require(data.length >= 4, "Data not a valid calldata");
+        require(bytes4(data[:4]) == this.requestUpdateOwner.selector, "Data not requestUpdateOwner");
+        (Forge forge, bytes memory name) = abi.decode(data[4:], (Forge, bytes));
+        _requestUpdateOwner(forge, name, amount);
```
xmxanuel, line 256: We let the user here pass an `amount` unequal to `fee`?  Maybe we should `revert` if the fee is not sufficient or use the `fee` variable.
On the other hand currently, the funds locked in the contract can be used by just calling `requestUpdateOwner`.

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+        // `driverId (32 bits) | forgeId (8 bits) | nameEncoded (216 bits)`
+        userId = (userId << 216) | nameEncoded;
+    }
+
+    /// @notice Updates the AnyApi operator configuration. Callable only by the admin.
+    /// @param operator The new address of the AnyApi operator.
+    /// @param jobId The new AnyApi job ID used for requesting user owner updates.
+    /// @param defaultFee The new fee in Link for each user owner
+    /// update request when the driver is covering the cost.
+    function updateAnyApiOperator(OperatorInterface operator, bytes32 jobId, uint96 defaultFee)
```
xmxanuel, line 173: We currently don't check of `pending` requests here. This means if the `updateAnyApiOperator` happens between request and `updateOwnerByAnyApi`. The transaction would revert.
Maybe not a big problem but i want to mention it. 

For me the naming is a bit confusion and could be improved `updateAnyApiOperator` `updateOwnerByAnyApi`. 
Completely, different things but overlapping terms.
Maybe:
```solidity
updateOwnerByAnyApi => fulfillUpdateOwner
```

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+        internal
+        returns (uint256 userId)
+    {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        address operator = address(storageRef.operator);
+        require(operator != address(0), "Operator address not set");
+        uint256 nonce = storageRef.nonce++;
+        bytes32 requestId = keccak256(abi.encodePacked(this, nonce));
+        userId = calcUserId(forge, name);
+        storageRef.requestedUpdates[requestId] = userId;
```
xmxanuel, line 275: Wondering if we should track the pending requests with a counter. Which is increased here and decreased after request fulfillment.

xmxanuel (commented):
Nice work. I added a few ideas and comments.

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+    /// It must be a valid ABI-encoded calldata for `requestUpdateOwner`.
+    /// The call parameters will be used the same way as when calling `requestUpdateOwner`,
+    /// to determine which user's ownership update is requested.
+    function onTokenTransfer(address, /* sender */ uint256 amount, bytes calldata data)
+        public
+        whenNotPaused
+    {
+        require(msg.sender == address(linkToken), "Callable only by the Link token");
+        require(data.length >= 4, "Data not a valid calldata");
+        require(bytes4(data[:4]) == this.requestUpdateOwner.selector, "Data not requestUpdateOwner");
```
xmxanuel, line 254: Mhm, as far as I know, the `data` is optional and it is up to us, how we want to define it.
We could just assume it is only the `data` and not the function signature. 
I think if we want this additional safety check, we should call the corresponding function and not the internal one. 
It would also solve the problem of the fee.

src/NFTDriver.sol
```diff
@@ -11,7 +11,6 @@ import {
 } from "./DripsHub.sol";
 import {Managed} from "./Managed.sol";
 import {Context, ERC2771Context} from "openzeppelin-contracts/metatx/ERC2771Context.sol";
-import {StorageSlot} from "openzeppelin-contracts/utils/StorageSlot.sol";
```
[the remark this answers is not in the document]
CodeSandwich: clean up.

src/RepoDriver.sol
CodeSandwich: Yes, accepting a custom fee is on purpose, it provides a fallback in case the fees change and we don't update the configuration. It's really hard to tell whether the fee is sufficient, there's no information available on-chain.

Just to be clear, this code is called after receiving `amount` Link tokens from the user, we're passing 100% of the received funds as a fee.

src/RepoDriver.sol
CodeSandwich: That's a good point, we could relatively easily clear the `requestedUpdates` mapping when updating the operator, e.g.
```diff
- mapping(bytes32 requestId => uint256 userId) requestedUpdates;
+ mapping(OperatorInterface operator => mapping(bytes32 requestId => uint256 userId)) requestedUpdates;
```
OTOH what would it protect from? If the operator is malicious, they will steal whatever identity they want anyway. I'm not even sure if removing the `delete storageRef.requestedUpdates[requestId];` line is worth having.

I'm fine with adding the `fulfill` prefix, it's a little verbose, but it's a clear counterpart to `request`. OTOH I'd keep the `ByAnyApi` part, it's meant to cleanly differentiate the API after the update to Chainlink Functions. But maybe it's fine to have clashing names, what bad could happen, and the updated function will probably have different parameters anyway :thinking:

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+    /// @param name The name of the repository.
+    /// @return url The built URL.
+    function _requestUrl(Forge forge, bytes memory name)
+        internal
+        pure
+        returns (string memory url)
+    {
+        if (forge == Forge.GitHub) {
+            return string.concat(
+                "https://raw.githubusercontent.com/", string(name), "/HEAD/FUNDING.json"
```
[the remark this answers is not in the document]
CodeSandwich: Other forges can't be added without an update because it involves not only a new URL but also a new user ID calculation logic. We'd need to implement something like a diamond or a registry of contracts per forge ID to allow adding new forges without touching the core.

No URL updates probably will be necessary in years, and we're already planning switching to Chainlink functions. All URLs and paths will be outsourced to JS code anyway, the admins will probably be able to switch it to a different one when there's a need for a URL change. I'd keep it simple for now.

src/RepoDriver.sol
CodeSandwich: A counter like `mapping(uint256 userId => uint256) numOfPendingRequests`? What uses would it have?

src/RepoDriver.sol
xmxanuel: We can also keep it like that but it resulted in this `helper variable` here.
Maybe you can do it `assembly` and use the enum for it (for generating the hash). It would also simplify these `if` statements.

src/RepoDriver.sol
xmxanuel: My assumption here is if `fee >= defaultFee` it should be sufficient. However, we could also allow it to pass 100%. 
Then, it could be used in case the configured `defaultFee` is not sufficient anymore.

src/RepoDriver.sol
xmxanuel: It would just mean that some pending requests will fail. We could check this with a counter and only allow a change if no pending requests.  However, then this transaction could be the frontrunner again to block an update.

src/RepoDriver.sol
```diff
@@ -0,0 +1,530 @@
...
+    function updateOwnerByAnyApi(bytes32 requestId, bytes calldata ownerRaw) public whenNotPaused {
+        RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        require(msg.sender == address(storageRef.operator), "Callable only by the operator");
+        uint256 userId = storageRef.requestedUpdates[requestId];
+        require(userId != 0, "Unknown request ID");
+        delete storageRef.requestedUpdates[requestId];
+        require(ownerRaw.length == 20, "Invalid owner length");
+        address owner = address(bytes20(ownerRaw));
+        _repoDriverStorage().userOwners[userId] = owner;
+        emit OwnerUpdated(userId, owner);
```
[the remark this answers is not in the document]
xmxanuel: > If we don't emit it, it's still available by putting together the AnyApiOperatorUpdated events and OwnerUpdated chronologically, Subgraph handles patterns like this gracefully.

Okay then it should be fine.

src/RepoDriver.sol
xmxanuel: I thought more of a global one. :thinking:

src/RepoDriver.sol
CodeSandwich: If we drop the signature, we'll lose any possibility of adding other behaviors on receiving tokens, which makes me slightly uncomfortable. Having a full calldata of an existing public function makes it convenient to build the payload, there are tools both in Solidity and in JS. Passing a full calldata payload is also a normal thing in the ecosystem, e.g. the operator does that.

The fee parameter behavior is intended, we're discussing it above.

src/RepoDriver.sol
CodeSandwich: I don't get the idea, could you paste the (pseudo) code? I agree that the `if`s here aren't clever, my goal was to make the code legible. If I wanted to make it clever, I'd probably do a bit shifting and a binary flag for name being long, but that wouldn't be an obvious approach.

src/RepoDriver.sol
xmxanuel: Yeah, that's right. 
My point was checking for `this.functionA.selector` and then calling `functionB`. However, maybe i should see it more like calling. `function_A` :).

I guess it is okay.

src/RepoDriver.sol
CodeSandwich: The default fee can also be too high, which would force our users to pay unreasonable fees unless we lower the default fee. I think that putting a hard limit on the range of fees is an unnecessary complication, it maybe protects from typos in the transferred amounts, but it brings weird restrictions to the API. If the user wants to pay X Link for the verification, let them. We may put some sanity checks in the app to avoid stupid mistakes.

src/RepoDriver.sol
CodeSandwich: Oh yes, and if there was a single failed request for the repo in the past, e.g. when there was no `FUNDING.json` file yet, it'd be forever blocked.

src/RepoDriver.sol
xmxanuel: Fixed it. Tests are passing.

xmxanuel: Ah, I misunderstood the function.

CodeSandwich: The problem is that `shr(8, forge)` always returns `0`, `forge` is an 8-bit number. I'm quite certain that this code doesn't need assembly though :D

xmxanuel: This would only work if the enum contains explicit all 4 forgeIds.

```solidity
enum Forge {
    GitHub,
    GithubLong,
    GitLab
    GitLabLong
}
```

xmxanuel: > The problem is that shr(8, forge) always returns 0, forge is an 8-bit number. I'm quite certain that this code doesn't need assembly though :D
Definitely.  Just playing with ideas.

PR #301, opened 2023-05-16
Split DripsHub balance into drips and splits

CodeSandwich:


xmxanuel (approved):
looks good, great work. I couldn't find any way to exploit it. 
It was the right decision to add at least this separation. :rocket:

PR #304, opened 2023-06-12
Add RepoDriver initializer

CodeSandwich:
It's needed to make deployment possible without requiring the owner of the contract to do anything.

Additionally, support for unknown chains is added, it's needed to enable testing on other chains, e.g. Anvil or Mumbai.

### DIFF
--- a/src/RepoDriver.sol
+++ b/src/RepoDriver.sol
@@ -54,7 +54,7 @@ contract RepoDriver is ERC677ReceiverInterface, ERC2771Context, Managed {
     /// @notice Emitted when the AnyApi operator configuration is updated.
     /// @param operator The new address of the AnyApi operator.
     /// @param jobId The new AnyApi job ID used for requesting user owner updates.
-    /// @param defaultFee The new fee in Link for each user owner
+    /// @param defaultFee The new fee in Link for each user owner.
     /// update request when the driver is covering the cost.
     event AnyApiOperatorUpdated(
         OperatorInterface indexed operator, bytes32 indexed jobId, uint96 defaultFee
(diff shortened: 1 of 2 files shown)


src/RepoDriver.sol
```diff
@@ -78,13 +81,15 @@ contract RepoDriver is ERC677ReceiverInterface, ERC2771Context, Managed {
         mapping(bytes32 requestId => uint256 userId) requestedUpdates;
         /// @notice The new address of the AnyApi operator.
         OperatorInterface operator;
-        /// @notice The fee in Link for each user owner
+        /// @notice The fee in Link for each user owner.
```
xmxanuel, line 84: I think one of the things I admire in your work is your passion for detail.
(Assuming this is not some prettier library)

src/RepoDriver.sol
```diff
@@ -161,17 +165,43 @@ contract RepoDriver is ERC677ReceiverInterface, ERC2771Context, Managed {
...
+    /// @notice Updates the AnyApi operator configuration. Callable only by the admin.
+    /// @param operator The new address of the AnyApi operator.
+    /// @param jobId The new AnyApi job ID used for requesting user owner updates.
+    /// @param defaultFee The new fee in Link for each user owner.
+    /// update request when the driver is covering the cost.
+    function _updateAnyApiOperator(OperatorInterface operator, bytes32 jobId, uint96 defaultFee)
+        internal
     {
         RepoDriverAnyApiStorage storage storageRef = _repoDriverAnyApiStorage();
+        storageRef.isInitialized = true;
```
xmxanuel, line 204: It would be cheaper for a regular update call if this is done in the `initializeAnyApiOperator`. Since the boolean slot is shared with the `nonce` which is not updated here. 

A bit more expensive and not so clean for the init. Reading the slot twice.

xmxanuel (approved):
looks good. :+1:

PR #305, opened 2023-06-12
Add deployer modules

CodeSandwich:
Adds:
- Deployment of RepoDriver.
- Deterministic contract addresses. The root of all addresses is the deployer's wallet address, which then is used to deploy DripsDeployer using a standard CREATE2 factory, which then uses CREATE3 to deploy all the rest. The bytecodes and the order of deployment don't matter. The only bytecode that needs to be preserved is that of DripsDeployer, it's a fairly simple contract, so it's probably not worth the effort to make its deployment use the modified CREATE3 pattern, we probably can assume that we won't need a different implementation for future deployments on other chains.
- Deployer modules are independent, they don't need to be deployed in any particular order (unless they explicitly depend on each other), can be deployed in batches, can be deployed later, and may be skipped and never deployed at all. Driver IDs 0-99 are reserved for the current and future modules.
- The Shanghai fork bytecode size limitation is bypassed, the deployed bytecodes live in regular calldata and aren't passed all together in a single init code anymore. The Linux's limitation of 128KB per parameter passed into programs hit hard, the workaround is to break up the list of deployed modules into multiple smaller lists, so each parameter passed into `forge send` can fit in the 128KB limit.

More comments and more tests are needed, but that can be added later.


src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+    bytes32 internal immutable _dripsModuleSalt = "Drips";
+
+    function _dripsModule() internal view returns (DripsModule) {
+        address module = _moduleAddress(_dripsModuleSalt);
+        require(Address.isContract(module), "Drips module not deployed");
+        return DripsModule(module);
+    }
+}
+
+contract DripsModule is DripsDependentModule, ProxyDeployerModule {
```
xmxanuel, line 150:
```suggestion
contract DripsModule is ProxyDeployerModule {
```

xmxanuel: I don't think the `DripsModule` needs this dependency

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+        bytes memory proxyInitCode = abi.encodePacked(type(ManagedProxy).creationCode, proxyArgs());
+        Create3.deploy(0, proxySalt, proxyInitCode);
+    }
+}
+
+abstract contract DripsDependentModule is BaseModule {
+    // slither-disable-next-line naming-convention
+    bytes32 internal immutable _dripsModuleSalt = "Drips";
+
+    function _dripsModule() internal view returns (DripsModule) {
```
xmxanuel, line 143: It is kind of a wild pattern and elegant pattern. A module which needs to get driverID from the DripsModule doesn't need to store the actual address it only requires the salt.

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+        proxyAdmin = proxyAdmin_;
+        // slither-disable-next-line too-many-digits
+        bytes memory proxyInitCode = abi.encodePacked(type(ManagedProxy).creationCode, proxyArgs());
+        Create3.deploy(0, proxySalt, proxyInitCode);
+    }
+}
+
+abstract contract DripsDependentModule is BaseModule {
+    // slither-disable-next-line naming-convention
+    bytes32 internal immutable _dripsModuleSalt = "Drips";
```
xmxanuel, line 141: Maybe always with a pre-fix?
```suggestion
    bytes32 internal immutable _dripsModuleSalt = "drips-contracts.drips";
```

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+    bytes32 public immutable proxySalt = "proxy";
+    address public proxyAdmin;
+    address public logic;
+
+    function proxy() public view returns (address) {
+        return Create3.computeAddress(proxySalt);
+    }
+
+    function proxyArgs() public view returns (bytes memory) {
+        return abi.encode(logic, proxyAdmin);
```
xmxanuel, line 115: Maybe adding a sanity check here?
```solidity
require(logic != address(0) && proxyAdmin != address(0));
```

Because it requires the inherited contract to first set this variable before `_deployProxy` can be called.

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+    constructor(address dripsDeployer_, bytes32 moduleSalt_) {
+        require(
+            address(this) == Create3.computeAddress(moduleSalt_, dripsDeployer_),
+            "Invalid module deployment salt"
+        );
+        dripsDeployer = dripsDeployer_;
+        moduleSalt = moduleSalt_;
+    }
+
+    function args() public view virtual returns (bytes memory);
```
xmxanuel, line 79: What is the point of `args()`? It is currently not used. Is the purpose just on-chain documentation of used parameters for actual deployment? 

I get the point of `deploymentArgs` and logicArgs` which are used.

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+    bytes32 internal immutable _callerModuleSalt = "Caller";
+
+    function _callerModule() internal view returns (CallerModule) {
+        address module = _moduleAddress(_callerModuleSalt);
+        require(Address.isContract(module), "Caller module not deployed");
+        return CallerModule(module);
+    }
+}
+
+contract CallerModule is ContractDeployerModule, CallerDependentModule {
```
xmxanuel, line 198:
```suggestion
contract CallerModule is ContractDeployerModule {
```

Same here

src/DripsDeployer.sol
xmxanuel: Ah, I see you not using the method but you need the `_dripsModuleSalt` from the `DripsDependentModule`

src/DripsDeployer.sol
xmxanuel: It is needed for the `salt`.

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// @notice Creates a contract under a deterministic addresses
+/// derived only from the deployer's address and the salt.
+/// The deployment is a two-step process, first, a proxy is deployed using CREATE2 with
+/// the given salt, and then it's called with the bytecode of the deployed contract,
+/// the proxy uses it to deploy the contract using regular CREATE.
+/// If the deployed contract's constructor reverts, the proxy also reverts.
+/// If the proxy call has a non-zero value, it's passed to the deployed contract's constructor.
+/// Based on the bytecode from https://github.com/0xsequence/create3.
```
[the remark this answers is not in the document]
xmxanuel: I still need to look into it to understand the differences.

xmxanuel (approved):
Nice work. In the beginning, I thought it is a bit over-engineered but now after looking deeper into it. It makes sense. :+1: 

However, we still depend on the private key of the initial owner. If I understand correctly.
If this one gets lost, we can't deploy on a new network with the same addresses. 
(Not solvable as long as we have any form of governance)

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+        return abi.encode(initialOwner);
+    }
+
+    constructor(address initialOwner_) {
+        // slither-disable-next-line missing-zero-check
+        initialOwner = initialOwner_;
+        _transferOwnership(initialOwner);
+    }
+
+    function deployModules(
```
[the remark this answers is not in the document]
CodeSandwich: It's 4 lists of modules. Each list must is passed into `cast call` as a separate argument, and Linux by default puts a limit of 128KB per program argument. Because the modules must be hex-encoded, this limits each modules list to 64KB of bytecode. If we did the sane thing and accepted only 1 modules list, we wouldn't be able to do the deployment in a single transaction.

src/DripsDeployer.sol
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+abstract contract BaseModule {
+    address public immutable dripsDeployer;
+    bytes32 public immutable moduleSalt;
+
+    constructor(address dripsDeployer_, bytes32 moduleSalt_) {
+        require(
+            address(this) == Create3.computeAddress(moduleSalt_, dripsDeployer_),
```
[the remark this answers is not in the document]
CodeSandwich: In this case `msg.sender` is the `Create3` proxy, not `DripsDeployer`. There's no way to look up 2 layers of calls, we'd need to modify `Create3` to somehow append `msg.sender` after the init code and then extract it, it's much easier to just pass it around manually.

src/DripsDeployer.sol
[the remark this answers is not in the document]
CodeSandwich: It's the same. There's 1 implementation of `Create3` passed around that everybody adds to their libraries. I've just linked the (IIUC) the original one.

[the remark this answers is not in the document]
CodeSandwich: If you want weird trivia and a piece of a lovely wasted time, I wrote a modified `Create3` that can be used directly by EOAs: https://gist.github.com/CodeSandwich/3022c63e3660a8f1b2dee2a235687a88 Regular `Create3` requires 2 transactions, first deploying the proxy and then calling it with the deployed init code. This is unusable for EOAs, because you may be front-run. The modified `Create3` stores its own deployer and lets only that address do the deployment, enabling bypassing the singleton factory and simplifying the address deriving chain. There's only one problem, `CREATE2` exists only as an opcode, you can't make a `CREATE2` transaction :rofl:

src/DripsDeployer.sol
CodeSandwich: In a way it already is prefixed, the default seed for the `DripsDeployer` in the singleton factory is `DripsDeployer`, so it's already in the address deployment chain. I don't think that we want to deploy other protocols using that factory, so prefixing it is probably an overkill. Having the salt matching the contract name simplifies the deployment script too.

src/DripsDeployer.sol
CodeSandwich: All the `...args` functions are used for contract verification. I don't quite understand why, but Etherscan requires filing not only the source code, but also the constructor arguments, for whatever reason it can't extract them from the initcode and the constructor source code. We can patch together the arguments in the deployment script or we can do it in the contract. Doing this in the contract is closer to the source of truth, and it also nicely documents the deployment on-chain.

src/DripsDeployer.sol
CodeSandwich: Yes

src/DripsDeployer.sol
CodeSandwich: `_deployProxy` sets those variables, so they can't be not set. The worst that could happen is not calling `_deployProxy`, but that's tricky to catch without explicity calling `_verifyDeployProxyCalled` or something similar in the module. It'd be optimal if `_deployProxy` was just a constructor, because the constructor would require it to be called, but then it'd be difficult to do any preparation before deploying the proxy. Adding the safety check would only generate a nicer error in the script when it called `proxyArgs` on a badly written module. We probably should add safety checks for `logic()` and `proxyAdmin()` then too. I feel that it's not worth the effort, how about you?

scripts/deploy.sh
```diff
@@ -10,117 +10,283 @@ print_title() {
...
+}
+
+# Args: module name
+module_address() {
+    local SALT="$(cast --format-bytes32-string "$1")"
+    cast call "$DRIPS_DEPLOYER" "moduleAddress(bytes32)(address)" "$SALT"
+}
+
+# Args: module address, field name, field type
+query() {
```
xmxanuel, line 45: I like this helper functions

PR #306, opened 2023-07-03
Use CREATE3 factory

CodeSandwich:
This makes the deployment more deterministic and no longer dependent on the `DripsDeployer` bytecode. Now we will be able to compile it using a future compiler with new features and still get the same addresses. The address derivation path is now simplified too, there's a single factory that may be queried repeatedly to recreate the whole path from the deployer's wallet address to each contract address, only using known salts.


scripts/deploy.sh
```diff
@@ -101,35 +99,82 @@ verify_proxy_deployer_module() {
...
-    print_title "Sending funds to the address deploying CREATE2 deployer"
+deploy_deterministic_deployer() {
+    print_title "Sending funds to the address deploying deterministic deployer"
     # Taken from https://github.com/Arachnid/deterministic-deployment-proxy
     cast send $WALLET_ARGS --value "0.01ether" "0x3fAB184622Dc19b6109349B94811493BF2a45362"
 
-    print_title "Deploying the CREATE2 deployer"
+    print_title "Deploying deterministic deployer"
     # Taken from https://github.com/Arachnid/deterministic-deployment-proxy
     cast publish "0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7ffffffffffffffff\
```
xmxanuel, line 109: Wondering if we can make this not hard-coded. I just compared it with the bytecode in the readme.

src/DripsDeployer.sol
```diff
@@ -319,94 +315,51 @@ contract RepoDriverModule is CallerDependentModule, DriverModule(3) {
...
-    /// @return proxy The proxy's address.
-    /// @return addr The deployed contract's address.
-    function _computeAddress(bytes32 salt, address deployer)
-        private
-        pure
-        returns (address proxy, address addr)
-    {
-        proxy = Create2.computeAddress(salt, keccak256(PROXY_CREATION_CODE), deployer);
-        addr = address(uint160(uint256(keccak256(abi.encodePacked(hex"d694", proxy, hex"01")))));
+        _CREATE3_FACTORY.deploy{value: amount}(salt, creationCode);
```
xmxanuel, line 332: One question with this pattern: 
Could an attacker not use the CREATE3_FACTORY and deploy another contract with the same `salt` before Drips is deployed?

xmxanuel: How is that prevented?

scripts/deploy.sh
CodeSandwich: We probably need to have SOMETHING hardcoded. This is a pre-signed transaction payload, if we wanted to build it ourselves, we'd need to build a transaction with a hardcoded signature and bytecode. If we wanted to build the bytecode, we'd need to copy the YUL source and compile it with a specific version of solc. IMO it's not worth it, having a single, easily verifiable blob is probably the cleanest approach.

src/DripsDeployer.sol
CodeSandwich: The attacker can do that, but the contract would be deployed under a different address: https://github.com/ZeframLou/create3-factory/blob/06ec0ff36d41853dcd4399fbe2127aef801c4077/src/CREATE3Factory.sol#L21

xmxanuel: Nice, this makes sense.

src/DripsDeployer.sol
```diff
@@ -319,94 +315,51 @@ contract RepoDriverModule is CallerDependentModule, DriverModule(3) {
...
-        hex"3d5260086018f3";
-
-    /// @notice Deploys a contract under a deterministic address.
+/// @notice Deploys contracts using CREATE3.
+/// Each deployer has its own namespace for deployed addresses.
+library Create3Factory {
+    /// @notice The CREATE3 factory address.
+    /// It's always the same, see `deploy_create3_factory` in the deployment script.
+    ICreate3Factory private constant _CREATE3_FACTORY =
+        ICreate3Factory(0x6aA3D87e99286946161dCA02B97C5806fC5eD46F);
```
xmxanuel, line 324: Okay, so this wrapper library is a trick that you only need to hardcode the address  once, correct.

xmxanuel (approved):
looks good

PR #308, opened 2023-07-04
Switch from assetId to erc20

CodeSandwich:


xmxanuel (approved):
looks good. I like the change. I think the `uint256 assetId` was a bit over-engineered. 
I also searched the `src` folder for assetId etc, looks fine.

PR #309, opened 2023-07-04
Upgrades

CodeSandwich:


src/Caller.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-only
-pragma solidity ^0.8.19;
```
xmxanuel, line 2: Solidity 0.8.20 is the version where they adopted the `push0` opcode. 

I assume the nodes of the testnets are already upgraded.

However, it might Drips not combability with other EVM based chains.

xmxanuel: I guess it is OK to not care about this much because we are Ethereum based and it will make things cheaper.

CodeSandwich: It shouldn't be a problem, Polygon is either working on it or it's already working, I'm not sure. For other chains we can do some adjustments for the deployment, switch back to `london` to avoid `PUSH0` and decrease `optimizer_runs`. Now we can do that freely because we're using the CREATE3 factory pattern for the deployment :smiley:

PR #316, opened 2023-09-11
Add DripsDataStore

CodeSandwich:


src/dataStore/DripsDataStore.sol
```diff
@@ -0,0 +1,240 @@
...
+            emit AccountMetadataStored(hash);
+        }
+    }
+
+    /// @notice Checks if the list of account metadata is stored.
+    /// @param hash The hash of the list, see `storeSplits`.
+    /// @return isStored True if the list of account metadata is stored.
+    /// If `hash` is `EMPTY_HASH`, which is `0`, always returns true.
+    function isAccountMetadataStored(bytes32 hash) public view returns (bool isStored) {
+        if (hash == EMPTY_HASH) return true;
```
xmxanuel, line 134:
```suggestion
        if (hash == EMPTY_HASH) return false;
```
I think it should be `false` here. It would work without the `if` as well.

src/dataStore/DripsDataStore.sol
```diff
@@ -0,0 +1,240 @@
...
+            emit StreamsHistoryStored(hash);
+        }
+    }
+
+    /// @notice Checks if the list of streams history entries is stored.
+    /// @param hash The hash of the list, see `storeSplits`.
+    /// @return isStored True if the list of streams history entries is stored.
+    /// If `hash` is `EMPTY_HASH`, which is `0`, always returns true.
+    function isStreamsHistoryStored(bytes32 hash) public view returns (bool isStored) {
+        if (hash == EMPTY_HASH) return true;
```
xmxanuel, line 174: same here.
```suggestion
        if (hash == EMPTY_HASH) return false;
```

src/dataStore/DripsDataStore.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    /// @notice Loads the stored data and returns it.
+    /// This function never returns, it stops the execution in the current context,
+    /// and the loaded data is returned as the external function's return data.
+    /// There is no validation, so the loaded data must be ABI-encoded and must
+    /// match exactly the returned data type of an external function that calls this function.
+    /// E.g. if a pointer stores ABI-encoded `uint[]`, ONLY an external function
+    /// returning `uint[]` can call `_loadAndReturn` for that pointer.
+    /// This is cheaper than loading data, ABI-decoding it, and then re-ABI-encoding to return it.
+    /// @param pointer The contract to load data from, see `_storeCalldata`. Reverts if `0`.
```
xmxanuel, line 230: okay, i think i understand the approach.

src/dataStore/DripsDataStore.sol
```diff
@@ -0,0 +1,240 @@
...
+    /// Reverts if the list with the given hash isn't stored, see `isSplitsStored`.
+    /// @param hash The hash of the list, see `storeSplits`.
+    /// @return accountMetadata The loaded list of account metadata.
+    function loadAccountMetadata(bytes32 hash)
+        external
+        view
+        returns (AccountMetadata[] memory accountMetadata)
+    {
+        if (hash == EMPTY_HASH) return accountMetadata;
+        _loadAndReturn(_accountMetadataPointers[hash]);
```
xmxanuel, line 148: okay we don't need a `return _loadAndReturn(...);` here because 
```
/// and the loaded data is returned as the external function's return data.
````

File: src/dataStore/DripsDataStore.sol [hunk 5]
```diff
@@ -0,0 +1,240 @@
...
+    /// returning `uint[]` can call `_loadAndReturn` for that pointer.
+    /// This is cheaper than loading data, ABI-decoding it, and then re-ABI-encoding to return it.
+    /// @param pointer The contract to load data from, see `_storeCalldata`. Reverts if `0`.
+    function _loadAndReturn(address pointer) internal view {
+        require(pointer != address(0), "Requested data not in storage");
+        bytes memory data = pointer.code;
+        // slither-disable-next-line assembly
+        assembly ("memory-safe") {
+            // Skip the first byte, it's the STOP opcode prefix.
+            return(add(data, 33), sub(mload(data), 1))
```
xmxanuel, line 237:
```suggestion
            return(add(data, 33), sub(mload(0x40), 1))
```
You could also use the free memory pointer here `0x40`.  would work as well.

Maybe a bit more comments? Explaining that you skip the first slot of 32 bytes.

xmxanuel: Okay, so the `return` opcode has two params 
- starting offset
- length

Why does `data` work here? In my understanding the code is now in memory and `data` is a pointer to the start of the code.

xmxanuel (commented):
I added a few questions and I found one potential issue. First impressions looks good, will look a bit deeper. 

I assume you looked into https://github.com/0xsequence/sstore2/ but you decided against using it because the offer more flexiblity `start` and `end` position. Which we don't need in our case, correct?

src/dataStore/DripsDataStore.sol
xmxanuel: Ahh got it. The first positions of data in memory is the array length.

src/dataStore/DripsDataStore.sol
```diff
@@ -0,0 +1,240 @@
...
+    /// Storing an already stored list is valid, and it doesn't do anything.
+    /// @param streams The streams receivers list to store.
+    /// @return hash The hash of the stored list, that can be used to load it.
+    /// If `streams` is empty, it's `EMPTY_HASH`, which is `0`.
+    /// It's always equal to the streams hash in `Drips` of the stored list.
+    function storeStreams(StreamReceiver[] calldata streams) external returns (bytes32 hash) {
+        if (streams.length == 0) return EMPTY_HASH;
+        bytes memory data = abi.encode(streams);
+        hash = keccak256(data);
+        if (_streamsPointers[hash] == address(0)) {
```
xmxanuel, line 53: Okay, so basically we would allow a user to call the `storeStreams` twice with the same data.

I think it makes sense to not `revert` in this case. Why not allow it, we can skip the `_store` part. Theoretically, this could happen.

However, I would recommend to still emit the  `emit StreamsStored(hash);` in this case. 

In case someone is waiting for the `event`.

src/dataStore/DripsDataStore.sol
CodeSandwich: Returning `true` matches the behavior of `loadAccountMetadata`, which is not throwing and returning the requested data, which is an empty array. It also matches `storeAccountMetadata`, if you pass into it an empty array and get back a hash (in this case `EMPTY_HASH`), then you expect that `isAccountMetadataStored` will return for that hash `true`.

src/dataStore/DripsDataStore.sol
CodeSandwich: This part could be implemented in many ways, depending on how much assembly we want. I wanted to reduce and encapsulate it as much as possible, so it's a single step, carving out the relevant data from memory and returning it. The "optimal" approach would be probably to do `codecopy` in YUL skipping the first byte, dump it to memory wherever, even on `0x00`, and pass that to `return`. It wouldn't save much gas though, but would be even more confusing than what we have here.

src/dataStore/DripsDataStore.sol
CodeSandwich: Hmm, that's' a good point. OTOH the event has been emitted in the past, so the off-chain should be able to see it (unless they are working directly on a non-archival node on a forgetful network like Polygon). If we emit the event for double store, maybe we should also emit it with `EMPTY_HASH` when somebody calls `store...` for an empty array. Should we add an emit of each kind with `EMPTY_HASH` in the constructor too?

src/dataStore/DripsDataStore.sol
xmxanuel: This makes sense.  I think the current implementation is a good one.

xmxanuel (approved):
looks good

PR #320, opened 2023-09-20
Add DripsDataProxy

CodeSandwich:


src/dataStore/DripsDataProxy.sol
```diff
@@ -0,0 +1,168 @@
...
+/// Large parameters aren't passed in calldata, but instead their hashes are accepted,
+/// and the actual data is loaded from `DripsDataStore`.
+/// The data must be explicitly stored in `DripsDataStore` before its usage.
+///
+/// In some cases using the proxy is easier and cheaper than calling the driver directly.
+/// For example multisigs and governance contracts often need the whole calldata
+/// provided as a binary blob, which then sometimes needs to be stored on-chain.
+/// It's easier to build calldata consisting of a flat list of scalar parameters,
+/// and calldata is much smaller when large lists are substituted with their hashes.
+contract DripsDataProxy is Managed {
```
xmxanuel, line 18: Why making it `Managed` it is a contract with no state or `storage`?

src/dataStore/DripsDataProxy.sol
```diff
@@ -0,0 +1,168 @@
...
+    /// Only streams described by `streamsHistoryHash` will be squeezed.
+    /// If `streamsHistoryHash` entries have no receivers, they won't be squeezed.
+    /// @return amt The squeezed amount.
+    function squeezeStreams(
+        uint256 accountId,
+        IERC20 erc20,
+        uint256 senderId,
+        bytes32 historyStartHash,
+        bytes32 streamsHistoryHash
+    ) public whenNotPaused returns (uint128 amt) {
```
xmxanuel, line 58: Since the underlying functions have a `whenNotPaused` modifier why here again?

xmxanuel (approved):
looks good. Basically, this helper contract allows the usage of the public Drips functions with the `DataStorage`.

One idea could be to provide functions where the `calldata` can be passed but it is stored in data storage for future usage.

src/dataStore/DripsDataProxy.sol
CodeSandwich: `Managed` is mostly about adding upgrades and pausing to the contract. The storage helpers is a bonus for upgrading, but it doesn't need to be used, the compiler will probably just remove its logic from the bytecode.

src/dataStore/DripsDataProxy.sol
CodeSandwich: Because `Drips` will be paused only if the core protocol is broken. We need a finer-grained pausing too in case the proxy is broken. It's the same with the drivers, each of them can be paused separately in case they are broken, but the protocol itself is fine.

PR #342, opened 2024-02-26
Switch to solc 0.8.24, upgrade dependencies and enable via-ir

CodeSandwich:
Some changes were needed to make the code via-IR-compatible.

The behavior of the inline initialization (e.g. `uint256 internal immutable foo = 123;`) is different, it's no longer done before all the constructors are run, but only right before the constructor of the contract containing this variable, so it can't be used as a parameter of the parent contract, it's silently set to 0.

Via-IR has some problems with very large functions, it probably tries to inline them and then optimize, which it fails to do. This is why some tests needed to be prevented from inlining by making the complicated function calls external.

Via-IR assumes that `block.timestamp` never changes, so it sometimes replaces accessing variables known to be holding the timestamp with `block.timestamp` retrieval, which derailed some tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,4 +23,4 @@ jobs:
       - name: run Slither
         uses: crytic/slither-action@v0.3.0
         with:
-          slither-version: '0.9.5'
+          slither-version: '0.10.0'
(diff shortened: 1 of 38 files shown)


jo-elimu (approved):

xmxanuel:
>Via-IR assumes that block.timestamp never changes, so it sometimes replaces accessing variables known to be holding the timestamp with block.timestamp retrieval, which derailed some tests.

Interesting.

I started to read about it here:
https://github.com/foundry-rs/foundry/issues/1373

xmxanuel (approved):
looks good. 

One minor thing.  I would add to the PR name that it includes a solidity version upgrade.

PR #343, opened 2024-02-26
Remove the pauser role

CodeSandwich:
The pauser role isn't used in any way, it's unclear how we could use it without centralization, and the safety pausability is supposed to bring is questionable. The proposed admin is moved to a different storage slot, so there should be no proposed admin pending during an upgrade.


src/AddressDriver.sol
```diff
@@ -28,7 +28,7 @@ contract AddressDriver is DriverTransferUtils, Managed {
     /// `driverId (32 bits) | zeros (64 bits) | addr (160 bits)`.
     /// @param addr The address
     /// @return accountId The account ID
-    function calcAccountId(address addr) public view returns (uint256 accountId) {
+    function calcAccountId(address addr) public view onlyProxy returns (uint256 accountId) {
```
jo-elimu, line 31: This change didn't _replace_ any `whenNotPaused`, like in the examples below?

CodeSandwich: It didn't. Using `view` functions without the proxy can return confusing data, so if we want to use `onlyxProxy` at all, then it probably makes sense to secure all the functions, not only the state-modifying ones.

src/NFTDriver.sol
```diff
@@ -370,56 +372,88 @@ contract NFTDriver is ERC721Burnable, DriverTransferUtils, Managed {
...
+    function safeTransferFrom(address from, address to, uint256 tokenId, bytes memory data)
+        public
+        override
+        onlyProxy
+    {
+        super.safeTransferFrom(from, to, tokenId, data);
+    }
+
+    /// @inheritdoc ERC721Burnable
+    function burn(uint256 tokenId) public override onlyProxy {
```
xmxanuel, line 455: Interesting, that `burn` is not part of any interface and it is not required to consider it in the `supportsInterface`

src/Managed.sol
```diff
@@ -183,16 +91,6 @@ abstract contract Managed is UUPSUpgradeable {
...
-    function _managedStorage() internal view returns (ManagedStorage storage storageRef) {
-        bytes32 slot = _managedStorageSlot;
-        // slither-disable-next-line assembly
-        assembly {
-            storageRef.slot := slot
-        }
-    }
-
     /// @notice Authorizes the contract upgrade. See `UUPSUpgradeable` docs for more details.
     function _authorizeUpgrade(address /* newImplementation */ ) internal view override onlyAdmin {
```
xmxanuel, line 95: Following the PR description I thought you want to check here if the `proposedAdmin` equals `address(0)`?

CodeSandwich: No, no, the PR description states that the slot changes, so any pending admin address will forever stay in storage without ever being accessed, and from the upgrade forward a different slot will be used to store the pending admin address. I think that it's a good tradeoff for keeping the code clean and the storage layout legacy-free. There hasn't been a single admin upgrade ever performed, so we can probably avoid having a pending admin during the upgrade.

I think that we should accept `address(0)` because it's a nice way to explicitly cancel the admin change proposal, it's obvious and it doesn't require a new function.

src/NFTDriver.sol
CodeSandwich: Yeah, I guess that it isn't required by ERC-721 so it isn't covered. I suspect that ERC-165 is considered to be a failed standard, so it's avoided in OZ wherever it isn't explicitly required by other standards. I don't think that I have seen ERC-165 used anywhere yet, as in somebody actually calling `supportsInterface` and checking the result.

src/Managed.sol
```diff
@@ -122,54 +71,13 @@ abstract contract Managed is UUPSUpgradeable {
     /// @notice Sets the current admin of the contract and clears the proposed admin.
     /// @param newAdmin The admin address being set. Can be the zero address.
     function _updateAdmin(address newAdmin) internal {
-        emit AdminChanged(admin(), newAdmin);
-        _managedStorage().proposedAdmin = address(0);
+        _proposedAdmin().value = address(0);
         StorageSlot.getAddressSlot(_ADMIN_SLOT).value = newAdmin;
```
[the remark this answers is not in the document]
CodeSandwich: `_updateAdmin` is used in `renounceAdmin` too which sets the admin address to address `0` and `_changeAdmin` requires that the address is not `0`. `_changeAdmin` could be used in `acceptAdmin` only, but then there would be 2 paths doing the same things and scattered across multiple files, one of them with a `require` that can't be triggered (`acceptAdmin` sets the admin to `msg.sender`) and one without the `require` because it would always be triggered (`renounceAdmin` sets the admin to `0`).

src/NFTDriver.sol
```diff
@@ -370,56 +372,88 @@ contract NFTDriver is ERC721Burnable, DriverTransferUtils, Managed {
...
+
     /// @inheritdoc IERC721Metadata
-    function name() public pure override returns (string memory) {
+    function name() public view override onlyProxy returns (string memory) {
         return "Drips identity";
     }
 
     /// @inheritdoc IERC721Metadata
-    function symbol() public pure override returns (string memory) {
+    function symbol() public view override onlyProxy returns (string memory) {
```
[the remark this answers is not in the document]
CodeSandwich: Good catch! Restored to `pure` with no `onlyProxy` modifier. Also removed `onlyProxy` from `supportsInterface`, despite being only `view` it acts like `pure`, it only depends on the bytecode.

PR #344, opened 2024-02-28
Optimize gas usage

CodeSandwich:


src/Splits.sol
```diff
@@ -160,29 +152,31 @@ abstract contract Splits {
...
-        if (collectableAmt == 0) {
+        // Clear the bits outside of the mask
+        uint128 splittable = balance.splittable & _SPLITTABLE_MASK;
+        if (splittable == 0) {
             return (0, 0);
         }
-        balance.splittable = 0;
+        // Set the value of `splittable` to `0`,
+        // and the bits outside of the mask to `1` to keep the storage slot non-zero.
+        balance.splittable = ~_SPLITTABLE_MASK;
```
xmxanuel, line 162: Wouldn't it be cheaper to just make

```solidity
balance.splittable = 2**127;
```

src/Splits.sol
```diff
@@ -25,32 +25,23 @@ abstract contract Splits {
     uint32 internal constant _TOTAL_SPLITS_WEIGHT = 1_000_000;
     /// @notice The amount the contract can keep track of each ERC-20 token.
     // slither-disable-next-line unused-state
-    uint128 internal constant _MAX_SPLITS_BALANCE = type(uint128).max;
+    uint128 internal constant _MAX_SPLITS_BALANCE = _SPLITTABLE_MASK;
```
xmxanuel, line 28: It is not only a PR for gas optimization it limits the max split balance to `type(uint128).max/2`.

xmxanuel (approved):
looks good. 

One idea:
I think it would be nice if the bit operations could be hidden from the original logic. Encapsulated in functions, if no extra gas costs are produced.

src/Splits.sol
CodeSandwich: It's treated as a constant anyway, so it's all the same. I think that using `_SPLITTABLE_MASK` here is a bit cleaner than the magic number `2**127`.

src/Splits.sol
CodeSandwich: Yes, that's a good observation. It's not visible on the outside of the protocol though. We have a test in place checking that `_MAX_SPLITS_BALANCE` isn't smaller than the protocol-wide `MAX_TOTAL_BALANCE`.

PR #345, opened 2024-02-28
Make the splits weights uint256

CodeSandwich:
This makes the API simpler and more sane, it requires less effort and gas to parse and validate the calldata.

### DIFF
--- a/src/Splits.sol
+++ b/src/Splits.sol
@@ -115,11 +115,11 @@ abstract contract Splits {
     /// @return collectableAmt The amount made collectable for the account
     /// on top of what was collectable before.
     /// @return splitAmt The amount split to the account's splits receivers
-    function _splitResult(uint256 accountId, SplitsReceiver[] memory currReceivers, uint128 amount)
-        internal
-        view
-        returns (uint128 collectableAmt, uint128 splitAmt)
-    {
+    function _splitResult(
+        uint256 accountId,
+        SplitsReceiver[] calldata currReceivers,
+        uint128 amount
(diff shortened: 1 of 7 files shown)


xmxanuel (approved):
looks good.

src/Splits.sol
```diff
@@ -240,15 +244,16 @@ abstract contract Splits {
...
             require(receivers.length <= _MAX_SPLITS_RECEIVERS, "Too many splits receivers");
             uint256 totalWeight = 0;
             uint256 prevAccountId = 0;
             for (uint256 i = 0; i < receivers.length; i++) {
-                SplitsReceiver memory receiver = receivers[i];
-                uint32 weight = receiver.weight;
+                SplitsReceiver calldata receiver = receivers[i];
+                uint256 weight = receiver.weight;
                 require(weight != 0, "Splits receiver weight is zero");
+                if (weight > _TOTAL_SPLITS_WEIGHT) weight = _TOTAL_SPLITS_WEIGHT + 1;
```
xmxanuel, line 256: Okay, just to verify if I understood this correctly. 
```solidity
weight = _TOTAL_SPLITS_WEIGHT + 1;
```

This is a gas optimization. You want to operate with `unchecked` and still have a  the potential overflow protection.  Therefore you check if the `weight` is greater than the `_TOTAL_SPLITS_WEIGHT`. 

You add `+1` to ensure it will revert later with `totalWeight <= _TOTAL_SPLITS_WEIGHT`  an additional `require` would be more expensive.

CodeSandwich: Yes, exactly :+1:

PR #346, opened 2024-02-28
Make max end hints a custom type

CodeSandwich:
This makes the calldata simpler by replacing zero-padded 4-byte words with a full word. The API also becomes cleaner by reducing the number of parameters.

### DIFF
--- a/src/Streams.sol
+++ b/src/Streams.sol
@@ -144,6 +144,59 @@ library StreamConfigImpl {
     }
 }
 
+/// @notice The list of 8 hints for max end time calculation.
+/// They are constructed as a concatenation of 8 32-bit values:
+/// `the leftmost hint (32 bits) | ... | the rightmost hint (32 bits)
+type MaxEndHints is uint256;
+
+using MaxEndHintsImpl for MaxEndHints global;
+
+library MaxEndHintsImpl {
+    /// @notice Create a list of 8 zero value hints.
+    /// @return hints The list of hints.
+    // slither-disable-next-line dead-code
+    function create() internal pure returns (MaxEndHints hints) {
+        return MaxEndHints.wrap(0);
+    }
+
+    /// @notice Add a hint to the list of hints as the rightmost and remove the leftmost.
+    /// @param hints The list of hints.
+    /// @param hint The added hint.
+    /// @return newHints The modified list of hints.
+    // slither-disable-next-line dead-code
+    function push(MaxEndHints hints, uint32 hint) internal pure returns (MaxEndHints newHints) {
+        // `hints` has value:
+        // `leftmost hint (32 bits) | other hints (224 bits)`
+        // By bit shifting we get value:
+        // `other hints (224 bits) | zeros (32 bits)`
+        // By bit masking we get value:
+        // `other hints (224 bits) | rightmost hint (32 bits)`
+        return MaxEndHints.wrap((MaxEndHints.unwrap(hints) << 32) | hint);
+    }
+
+    /// @notice Remove and return the rightmost hint, and add the zero value hint as the leftmost.
+    /// @param hints The list of hints.
+    /// @return newHints The modified list of hints.
+    /// @return hint The removed hint.
+    function pop(MaxEndHints hints) internal pure returns (MaxEndHints newHints, uint32 hint) {
+        // `hints` has value:
+        // `other hints (224 bits) | rightmost hint (32 bits)`
+        // By bit shifting we get value:
+        // `zeros (32 bits) | other hints (224 bits)`
+        // By casting down we get value:
+        // `rightmost hint (32 bits)`
+        return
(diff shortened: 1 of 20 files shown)


src/Streams.sol
```diff
@@ -683,29 +736,29 @@ abstract contract Streams {
     /// @param newReceivers The list of the streams receivers of the account to be set.
     /// Must be sorted by the account IDs and then by the stream configurations,
     /// without identical elements and without 0 amtPerSecs.
-    /// @param maxEndHint1 An optional parameter allowing gas optimization, pass `0` to ignore it.
-    /// The first hint for finding the maximum end time when all streams stop due to funds
+    /// @param maxEndHints An optional parameter allowing gas optimization.
+    /// Pass a list of 8 zero value hints to ignore it, it's represented as an integer `0`.
```
xmxanuel, line 740:
```suggestion
    /// Pass `0` to ignore it. 
```
That's too complicated in my opinion `list of 8 zero value hints` for the reader who only wants to skip it.

src/Streams.sol
```diff
@@ -683,29 +736,29 @@ abstract contract Streams {
...
+    /// The list of hints for finding the maximum end time when all streams stop due to funds
     /// running out after the balance is updated and the new receivers list is applied.
     /// Hints have no effect on the results of calling this function, except potentially saving gas.
     /// Hints are Unix timestamps used as the starting points for binary search for the time
     /// when funds run out in the range of timestamps from the current block's to `2^32`.
-    /// Hints lower than the current timestamp are ignored.
-    /// You can provide zero, one or two hints. The order of hints doesn't matter.
+    /// Hints lower than the current timestamp including the zero value hints are ignored.
+    /// If you provide fewer than 8 non-zero value hints make them the rightmost values to save gas.
+    /// It's the best approach to make the most risky and precise hints the rightmost ones.
```
xmxanuel, line 748: I think it would good to mention specifically that there are no assumptions.
- no specific ordering of the hints is strictly required
  - only the already mentioned gas improvement with risky hints on the right
- all hints could be used to improve either higher bound or the lower bound 
- it is not required that hints come in pairs of two with a start and end
- provided list only improves gas usage and has no impact on the outcome of the transaction

I think a potentially confusing part is: why is it a list with multiple hints. 

Maybe a few strategies to illustrate the concept. Alternatively a link to some documentation

For example: Unkown block.timstamp of the new transactions together with streams which should stop when running out of funds:

- risky hint pair a `1 min` start/end range
- 5 minute hint pair start/end range
- 2 hours pair for the worst case

src/Streams.sol
```diff
@@ -811,19 +859,14 @@ abstract contract Streams {
...
-                enoughEnd = hint1;
-            } else {
-                notEnoughEnd = hint1;
-            }
-        }
-
-        if (hint2 > enoughEnd && hint2 < notEnoughEnd) {
-            if (_isBalanceEnough(balance, configs, configsLen, hint2)) {
-                enoughEnd = hint2;
+        while (maxEndHints.hasHints()) {
```
xmxanuel, line 862: Just to double check.
```solidity
uint256 enoughEnd = _currTimestamp();
```
Current timestamp will be always `enoughEnd`. This is an invariant of the system and can never be false.

src/Streams.sol
```diff
@@ -811,19 +859,14 @@ abstract contract Streams {
...
+            uint32 hint;
+            (maxEndHints, hint) = maxEndHints.pop();
+            if (hint <= enoughEnd || hint >= notEnoughEnd) continue;
+            if (_isBalanceEnough(balance, configs, configsLen, hint)) {
+                enoughEnd = hint;
```
xmxanuel, line 867: Logic makes sense to me :+1:

src/Streams.sol
```diff
@@ -144,6 +144,59 @@ library StreamConfigImpl {
...
+        // By bit masking we get value:
+        // `other hints (224 bits) | rightmost hint (32 bits)`
+        return MaxEndHints.wrap((MaxEndHints.unwrap(hints) << 32) | hint);
+    }
+
+    /// @notice Remove and return the rightmost hint, and add the zero value hint as the leftmost.
+    /// @param hints The list of hints.
+    /// @return newHints The modified list of hints.
+    /// @return hint The removed hint.
+    function pop(MaxEndHints hints) internal pure returns (MaxEndHints newHints, uint32 hint) {
```
xmxanuel, line 181: It makes sense to implement it with a `pop` instead of a `read` operation for a specific list index. (more costly)

xmxanuel (approved):
Looks good. I think it is a nice improvement. However, it requires a good understanding of the `hints` to use it. Which might be okay because it is an optional gas improvement parameter.

src/Streams.sol
CodeSandwich: Clarified.

src/Streams.sol
CodeSandwich: Clarified.

PR #347, opened 2024-02-28
Make Drips balances uint256

CodeSandwich:
This makes the balances type more compatible with the ERC-20 balances.


xmxanuel (approved):
looks good

PR #358, opened 2025-01-28
docker setup

efstajas:
Adds a Docker setup that builds an image containing an `anvil` testnet with Drips contracts and a test ERC-20 pre-deployed, plus a workflow for publishing it to Dockerhub.


CodeSandwich (approved):
Looks good, I don't see any issues :+1:

PR #359, opened 2025-01-30
Set up docker image builds

efstajas:


scripts/deploy-docker.sh
```diff
@@ -0,0 +1,23 @@
...
+
+mkdir anvil-data
+
+while ! grep -q "Listening on 0.0.0.0:8545" <(anvil --dump-state ./anvil-state.json --host 0.0.0.0); do
+  sleep 1
+done
+
+forge script -f localhost:8545 --broadcast --slow --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 script/LocalTestnet.sol:Deploy
+if [ $? -ne 0 ]; then
+  exit 1
```
CodeSandwich, line 13: Can we replace the manual error handling with the classic bash spell `set -euxo pipefail` at the top of the script, so it bubbles automatically?

efstajas: Done 👍

scripts/deploy-docker.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -euxo pipefail
+
+echo Deploying contracts to anvil...
+
+mkdir anvil-data
+
+while ! grep -q "Listening on 0.0.0.0:8545" <(anvil --dump-state ./anvil-state.json --host 0.0.0.0); do
```
CodeSandwich, line 8: I hope that you know what you're doing here because I certainly don't :laughing: Anvil is run in a subshell with output piped into grep looking for the log. When does grep return the exit code? On the first match? When anvil exits/crashes and closes the pipe? If there's a match, doesn't stopping grep kill the subshell? If there's no match, should we try again or abort because there's clearly a serious problem and trying forever is probably a mistake?

efstajas: `grep` exits `0` as soon as it finds the string, yes. before that it exits with `1`. So this waits repeatedly for 1s until the testnet is up. I'm not an expert at shell scripting at all, and pieced all this together from various pieces i found on the internet. But it seems to be working^^

I just replaced it with something that I think is a bit clearer.

It'd still run indefinitely if the testnet fails to start up until it eventually gets killed by the CI run timing out... but adding a timeout here is a bit over my head. how would you do it?

CodeSandwich: I like the new version very much :+1:

There's a `timeout` command, e.g. `timeout 10 grep ...` will return a non-zero code if `grep` doesn't end itself in 10 seconds. If it does, grep's code is returned, which will be 0 if there's a match or 1 if there isn't and the pipe is closed.

PR #365, opened 2026-09-21
Add fuzz tests for funds adding up when squeezing

BhariGowda:
This replaces #364, which could not be reopened after a force-push (GitHub returned UNPROCESSABLE with no further detail on the reopen mutation; see the comment on #364 for the full technical writeup).

The three conservation fuzz tests in `Streams.t.sol` (`testFundsStreamedToReceiversAddUp`, `testFundsStreamedToReceiversAddUpAfterStreamsUpdate` and `testFundsStreamedFromSendersAddUp`) all take funds out with `receiveStreams`. None of them squeeze, so squeezing was not covered by any of the funds adding up checks. It is the only path that writes negative `amtDeltas`, so it looked like the gap most worth closing.

This adds two tests asserting the same property those three assert, that squeezed plus withdrawn plus received equals the balance the sender started with.

`testFundsSqueezedAndReceivedAddUp` squeezes twice at independently fuzzed times, each up to three cycles long. The two squeezes can land in the same cycle or in different ones, and the second one can fall after the streams have already run out.

`testFundsSqueezedAcrossStreamsUpdatesAddUp` squeezes once, replaces the receivers list mid-stream, then squeezes both the old and the new list against a two entry history. This covers the `currCycleConfigs` slot mapping in `_squeezeStreamsResult` and stale `nextSqueezed` values being clamped up to the current cycle start.

### Why the receivers need their own sanitizer

Squeezing only reaches back to the start of the current cycle, so with `cycleSecs` at 10 the widest squeezable window is 9 seconds. `sanitizeReceivers` leaves `amtPerSec` as low as `minAmtPerSec`, which is `AMT_PER_SEC_MULTIPLIER / cycleSecs`, and leaves `start` anywhere in the `uint32` range. Over a 9 second window a receiver at that rate is owed less than one whole token, so `_streamedAmt` rounds it down to zero.

Sanitized that way both tests pass with a squeezed amount of exactly zero on every run. I measured it before fixing it: 0 of 500 runs of the first test and 3 of 5000 runs of the second squeezed anything at all. The identity still held, but only in its trivial form, so the tests would have asserted nothing that `testFundsStreamedToReceiversAddUp` does not already assert more cheaply, and they would have kept passing if squeezing were broken.

`sanitizeSqueezedReceivers` keeps the ordering and deduplication `sanitizeReceivers` does, and pins the rate between 1 and 16 tokens per second and the start into the test's own timespan. Measured over 50000 runs each:

| | before | after |
| --- | --- | --- |
| `testFundsSqueezedAndReceivedAddUp` | 0.0% | 69.0% |
| `testFundsSqueezedAcrossStreamsUpdatesAddUp` | 0.06% | 75.1% |

Median nonzero squeeze is 375 and 337. The runs that still squeeze zero are empty receiver lists (10.4%), zero balances (1.6%) and squeeze times landing exactly on a cycle boundary, which gives a zero length window. Those are all worth keeping, so I did not filter them out. Counting only runs with a non-empty receiver list and a nonzero balance, the first test squeezes a nonzero amount in 79.7% of runs.

### Receivers list length

The two tests cap the receivers list at 10 instead of the 100 the existing tests use. They squeeze every receiver in the list, which is quadratic in its length, and at 100 the two of them cost more than the existing three combined. At 10 they run at mean gas of 1.25M and 2.29M against 8.1M for `testFundsStreamedToReceiversAddUpAfterStreamsUpdate`. `_squeezedAmt` sums each receiver's amount independently, so the list length is not a dimension of the squeezing accounting.

No source changes. `forge test --deny warnings` is green at 380 tests, `forge fmt --check` passes and Slither reports 0 findings. Both new tests pass at 50000 runs, the count CI uses on pushes.


BhariGowda:
CI shows action_required, that's GitHub's approval gate for Actions on fork PRs, not a test failure. Same thing happened on #364 and #363 before it. Local run is green: forge test --deny warnings passes (380 tests), forge fmt --check is clean.

